PR #10, opened 2021-03-16
Fix: workflow release

AnaMariaMoga:
Problem: right now, the CD pipeline creates a new branch on every new tag. 

What we want: to create a new branch (with the old version .x) for documentation if the minor or major from version is changing.

Solution: added a script that takes the new version we release and the old version as args and checks if the minor or major changed and it creates a new branch accordingly.


.github/workflows/release.yml
```diff
@@ -27,12 +29,12 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          branch: release/${{ steps.get_version.outputs.VERSION }}
+      - name: Get the new version
+        id: new_version
+        run: echo ::set-output name=VERSION::$(git tag --list "v*" --sort=refname | tail -n 1)
+      - name: Get the old version
+        id: old_version
+        run: echo ::set-output name=VERSION::$(git tag --list "v*" --sort=refname | tail -n 2 | head -n 1)
```
mflorin, line 37: @AnaMariaMoga - what you want here is actually `--sort=version:refname` . `refname` only sorts lexicographically  . e.g if you tagged in order v1.2.3, v2.3.4, v12.3.4, v23.3.4 - refname will give you 
```
$ git tag --list "v*" --sort=refname
v1.2.3
v12.3.4
v2.3.4
v23.3.4
```

If you run it with `--sort=version:refname` you'll get
```
v1.2.3
v2.3.4
v12.3.4
v23.3.4
```

mflorin (commented):
`git tags --list "v*" --sort=refname` should be changed to `--sort=version:refname`

.github/workflows/release.yml
AnaMariaMoga: updated

mflorin (approved):

PR #13, opened 2021-03-17
Use environment variables as auth config fallback

avorima:
This makes the CLI less reliant on the config file and allows to use it without calling `login` first.

Fallback is only used when the config file is not found.


AnaMariaMoga (approved):
Approving this.

PR #14, opened 2021-03-22
Support bearer token auth

avorima:
This should be possible.

I can add another prompt to the login command for the token if needed.


AnaMariaMoga:
> This should be possible.
> 
> I can add another prompt to the login command for the token if needed.

I think it would be nice to update login command for the token (without flags) as well. And I am glad the token usage is optional.

AnaMariaMoga (approved):

PR #18, opened 2021-04-05
Feat: add images and snapshots commands

AnaMariaMoga:
* Adding images and snapshots commands 
* Adding tests for the new features added
* Adding examples and documentation for the new features added


AnaMariaMoga:
Merging this PR since it is adding commands on the same structure as before.

PR #55, opened 2021-06-03
Feat: update server [v6]

AnaMariaMoga:
Description: 
- update `server create` command to support also `CUBE` servers
- added `template` commands, as templates are needed for Cube Servers creation
- added `server token` commands
- added `server console` commands
- updated tests

Documentation will be updated in a separate PR.


pkg/resources/server_test.go
```diff
@@ -64,6 +64,30 @@ func TestNewServerService(t *testing.T) {
...
+	})
+	t.Run("get_token_server_error", func(t *testing.T) {
+		svc := getTestClient(t)
+		backupUnitSvc := NewServerService(svc.Get(), ctx)
+		_, _, err := backupUnitSvc.GetToken(testServerResourceVar, testServerResourceVar)
+		assert.Error(t, err)
+	})
+	t.Run("get_remote_console_server_error", func(t *testing.T) {
+		svc := getTestClient(t)
+		backupUnitSvc := NewServerService(svc.Get(), ctx)
```
mflorin, line 87: not a big deal, but keeping the same `backupUnitSvc` variable name (probably pasted from the backupunit tests) is a bit misleading :)

mflorin (approved):

pkg/resources/server_test.go
AnaMariaMoga: indeed! didn't notice this. Thanks a lot

PR #76, opened 2021-07-01
Feat/enhancement2

### LINKED ISSUE #35: Allow to provide account credentials via environment variables
Let ionosctl source its credentials for account authentication from environment variables as an alternative to the config file based approach that is already implemented.

This would mean that you would not have to store the credentials, potentially unencrypted in a config file but in a transient environment variables.

This use of environment variables then could be combined with a tool like [envdir](https://github.com/direnv/direnv) that can load environment variables on the dynamically on the fly when you enter a project folder.

In combination with a tool like [gopass](https://github.com/gopasspw/gopass) those credentials could even be stored encrypted and then only dynamically loaded, encrypted and store in a transient environment variable by envdir.

The naming of the environment variables should be aligned with environment variables that we might are already use in other IONOS Cloud SDKs or tools like Terraform/Ansible.

We also need to define an order of precedence for different ways/locations that ionosctl is looking for credentials - falling back from one to the next.

gabisavu98:
fixes #35


commands/login_test.go
```diff
@@ -3,6 +3,8 @@ package commands
 import (
 	"bufio"
 	"bytes"
+	ionoscloud "github.com/ionos-cloud/sdk-go/v5"
+	"os"
 	"testing"
 
 	"github.com/ionos-cloud/ionosctl/pkg/config"
```
AnaMariaMoga, line 10: In Go, it is recommended to separate remote imports like `ionoscloud "github.com/ionos-cloud/sdk-go/v5"` for the other ones.

pkg/config/config.go
```diff
@@ -3,20 +3,21 @@ package config
...
 	"encoding/json"
 	"errors"
+	"fmt"
+	sdk "github.com/ionos-cloud/sdk-go/v5"
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 
-	sdk "github.com/ionos-cloud/sdk-go/v5"
```
AnaMariaMoga, line 10: Some comment as above.

pkg/config/config.go
```diff
@@ -33,25 +34,47 @@ func getConfigHomeDir() string {
...
-	viper.SetConfigFile(viper.GetString(ArgConfig))
-	err := viper.ReadInConfig()
-	if err != nil {
-		return err
+	path := viper.GetString(ArgConfig)
+	if !filepath.IsAbs(path) {
+		path, _ = filepath.Abs(path)
 	}
-	return nil
+	//fmt.Println("path: " + path)
```
AnaMariaMoga, line 41: Clean-up :D

AnaMariaMoga (approved):
Good job overall, just small comments. I approve this, and after fixing the Windows tests - you can merge it.

commands/login_test.go
gabisavu98: done

pkg/config/config.go
gabisavu98: done

pkg/config/config.go
gabisavu98: done

README.md
```diff
@@ -79,6 +79,15 @@ from the folder where you cloned the ionosctl git.
...
 
+* Using environment variables
+If you have set the enviroment credentials, skip this section, if not, you can set them or follow the steps below for using config a config file:
+  
+You can set the enviroment variables using the next commands:
+```
+export IONOS_USERNAME="ionos username"
+export IONOS_PASSWORD="ionos password"
+```
+
```
AnaMariaMoga, line 90: After today;s demo, I think it is important to add in here that environment variables have priority to config file. This applies also to the other README.md and to v6.

gabisavu98: changed

pkg/config/config.go
```diff
@@ -33,25 +35,62 @@ func getConfigHomeDir() string {
...
+	if system == "windows" {
+		if permNumber == int(666) {
+			viper.SetConfigFile(viper.GetString(ArgConfig))
+			err := viper.ReadInConfig()
+			if err != nil {
+				return err
+			}
+			return nil
+		} else {
+			fmt.Printf("perm: %v", permNumber)
```
AnaMariaMoga, line 63: Please remove the fmt.Printf in here and below, eventually, you can add it to the `errors.New("no permission for the config file, expected 600, got "+permNumber)` to be more explicit, but remaining this way will affect the json/text format printing of the tool.

gabisavu98: done

PR #80, opened 2021-07-09
now request command is showing target datacenters and their ids

gabisavu98:
ionosctl request now is showing which resources is impacting and theirs ids


AnaMariaMoga (commented):
Looks good to me, but we also need to update tests and documentation before merging, for both v5 and v6.

AnaMariaMoga (approved):

PR #86, opened 2021-07-27
Use Makefile syntax to define defaults

avorima:
Makefile uses `?=` to define a variable that is set if the variable on the left side is not defined. This also fixes the binary name when running `make build`.

Additionally, ensure that the docs directory exists before updating.

Before:

```sh
$ make
--- Building ionosctl via go build ---
built /home/mv/Projects/CORE/ionosctl/builds/ionosctl__
DONE
```

After:

```sh
make
--- Building ionosctl via go build ---
built /home/mv/Projects/CORE/ionosctl/builds/ionosctl_linux_amd64
DONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,11 +4,11 @@ export CGO_ENABLED = 0
 export GO111MODULE := on
 
 GOFILES_NOVENDOR=$(shell find . -type f -name '*.go' | grep -v vendor)
-GOOS=${GOOS:-$(shell go env GOOS)}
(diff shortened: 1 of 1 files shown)


avorima:
This also is broken is the release/v6 branch. Should I open an extra PR to v6 branch or do you cherry-pick commits?

AnaMariaMoga:
> This also is broken in the `release/v6` branch. Should I open an extra PR to v6 branch or do you cherry-pick commits?

Please open an extra PR for the `release/v6` branch as well.

PR #88, opened 2021-07-27
Fix wait for non-default API URLs

avorima:
Use a regex to extract the request ID instead of checking for the hard-coded default v5 API URL.

This also should fix printing in general for non-default API URLs, like staging.

### DIFF
--- a/pkg/utils/printer/printer_test.go
+++ b/pkg/utils/printer/printer_test.go
@@ -166,7 +166,7 @@ func TestPrinterResultJsonRequestId(t *testing.T) {
 			},
 		},
 	}
-	res.ApiResponse.Header.Add("location", "https://api.ionos.com/cloudapi/v5/requests/123456/status")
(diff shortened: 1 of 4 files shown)


pkg/utils/printer/printer_test.go
```diff
@@ -166,7 +166,7 @@ func TestPrinterResultJsonRequestId(t *testing.T) {
 			},
 		},
 	}
-	res.ApiResponse.Header.Add("location", "https://api.ionos.com/cloudapi/v5/requests/123456/status")
+	res.ApiResponse.Header.Add("location", "https://api.ionos.com/cloudapi/v6/requests/123456/status")
```
AnaMariaMoga, line 169: not a big deal, but please leave v5 version for `master` branch. This applies also to line 492.

AnaMariaMoga (approved):

pkg/utils/printer/printer_test.go
avorima: Done

PR #93, opened 2021-08-04
Store API URL in config file

avorima:
API URL is stored as `userdata.api-url` in the config file when `login` command is called. Subsequent calls read the URL from the config file.

The API URL can also be set via `IONOS_API_URL` environment variable.
This variable name is also used by the [terraform provider].

The sources for API URL have the following priority:

1. Explicitly set `--api-url` flag
2. Config file
3. Env var `IONOS_API_URL`
4. `--api-url` default value

Other changes:

- Fix login when the value of `--api-url` ends in `cloudapi/v5`
- Use config file over env when both `IONOS_USERNAME` and `IONOS_TOKEN` are not defined, was (`IONOS_USERNAME` **or** `IONOS_PASSWORD` before)

These changes make using `ionosctl` in staging environments a lot easier.

[terraform provider]: https://registry.terraform.io/providers/ionos-cloud/ionoscloud/latest/docs#usage

### DIFF
--- a/pkg/resources/v5/client.go
+++ b/pkg/resources/v5/client.go
@@ -2,7 +2,7 @@ package v5
 
 import (
 	"errors"
-	"fmt"
+	"strings"
 
 	"github.com/ionos-cloud/ionosctl/pkg/config"
 	ionoscloud "github.com/ionos-cloud/sdk-go/v5"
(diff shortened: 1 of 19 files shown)


pkg/testdata/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "api-url": "https://api.ionos.com/cloudapi/v5",
+  "userdata.api-url": "https://api.ionos.com/cloudapi/v5",
```
AnaMariaMoga, line 2: Not a big deal: but can you please update with `https://api.ionos.com` in here?

pkg/resources/v5/client.go
```diff
@@ -32,6 +32,9 @@ func NewClientService(name, pwd, token, hostUrl string) (ClientService, error) {
 	if hostUrl == "" {
 		return nil, errors.New("host-url incorrect")
 	}
+	if !strings.HasSuffix(hostUrl, config.DefaultV5BasePath) {
+		hostUrl += config.DefaultV5BasePath
+	}
```
AnaMariaMoga, line 37: :1st_place_medal:

AnaMariaMoga (approved):

PR #94, opened 2021-08-05
Store API URL in config file

avorima:
Cherry-pick of #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ You can set the environment variables using the next commands:
 export IONOS_USERNAME="ionos username"
 export IONOS_PASSWORD="ionos password"
 export IONOS_TOKEN="ionos token"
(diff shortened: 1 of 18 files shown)


avorima:
I noticed that there is an inconsistency in `NewClientService` between v5 and v6. The v6 branch uses the API URL as is, whereas v5 appends the default base path (if it isn't already present). Which behaviour is correct?

AnaMariaMoga:
For v6 the suffix with base path is a WIP, on a different PR, will be added today or tomorrow. This change was added mainly for supporting new services: example: `api.ionos.com/cloudapi/autoscaling/`

avorima:
Ok, I'll leave it as it is

AnaMariaMoga (approved):

PR #111, opened 2021-09-20
Feat/v5/add all flag

gabisavu98:
## Checklist

- [ ] PR Title containing `Feat`/`Fix`/`Doc`
- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] SONAR Cloud Scan - warnings checked
- [ ] Task updated (Jira/Github)


commands/cloudapi-v5/backupunit.go
```diff
@@ -182,6 +183,7 @@ Required values to run command:
 		return completer.BackupUnitsIds(os.Stderr), cobra.ShellCompDirectiveNoFileComp
 	})
 	deleteCmd.AddBoolFlag(config.ArgWaitForRequest, config.ArgWaitForRequestShort, config.DefaultWait, "Wait for the Request for BackupUnit deletion to be executed")
+	deleteCmd.AddBoolFlag(cloudapiv5.ArgAll, cloudapiv5.ArgAllShort, false, "delete all BackupUnits.")
```
AnaMariaMoga, line 186: Please use upper case at the beginning of flag's description. It is good practice to be in sync with the rest of the flags. This applies for ALL the places in this PR where the flag is added.

commands/cloudapi-v5/backupunit.go
```diff
@@ -278,19 +287,63 @@ func RunBackupUnitUpdate(c *core.CommandConfig) error {
...
-		return err
-	}
-	if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
-		return err
+	var resp *resources.Response
+	var err error
+	var backupUnits resources.BackupUnits
+	allFlag := viper.GetBool(core.GetFlagName(c.NS, cloudapiv5.ArgAll))
+	if allFlag {
+		fmt.Printf("Backup Unitts to be deleted:\n")
```
AnaMariaMoga, line 295: Why are we using `fmt` package instead of the logger/printer? This applies for ALL the places in the PR where `fmt` pkg is being used & added.

commands/cloudapi-v5/backupunit.go
```diff
@@ -278,19 +287,63 @@ func RunBackupUnitUpdate(c *core.CommandConfig) error {
...
+					}
+					if err != nil {
+						return err
+					}
+					if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
+						return err
+					}
+				}
+			}
+		}
```
AnaMariaMoga, line 331: Is there a way in which the "delete-all" functionality can be added in a separate function and that function to be called from here based on flag's value? Adding it like this seems to add a new functionality to a function. And alsoooo, a lot of code smells in the Sonar :)

AnaMariaMoga (commented):
A lot of times, the user can have resources which are dependent of other resource and that can block deleting them. In my opinion, the `all` flag should also cover removing or detaching all resources from a resource and should be implemented also for those commands. Maybe adding a task on this to keep track of this will help, since this PR has already been here for a while.

commands/cloudapi-v5/ipblock.go
```diff
@@ -232,22 +240,72 @@ func RunIpBlockUpdate(c *core.CommandConfig) error {
...
+				if err != nil {
+					return err
+				}
+				if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
+					return err
+				}
+			}
+		}
 	}
-	return c.Printer.Print(getIpBlockPrint(resp, c, nil))
```
AnaMariaMoga, line 250: Deleting this line, leaves the user without a feedback if the command has been executed and it looks something like this:
```
ionosctl ipblock delete -i 33f5029a-e802-4af4-95c1-a47df0445ab6 
Warning: Are you sure you want to delete ipblock (y/N) ? 
y
```
From the current behaviour: 
```
ionosctl ipblock delete -i 8caca11d-fb68-4160-843b-cf485f77fa6d 
Warning: Are you sure you want to delete ipblock (y/N) ? 
y
RequestId: b390b67b-a5c5-47a5-8709-979109bdba45
Status: Command ipblock delete has been successfully executed
```
In my opinion, it is better that the user has some feedback regarding a command, for both delete options (--all and --resource-id). What is your opinion on this? 

(This applies to ALL the places in the PR where this line has been deleted)

commands/cloudapi-v5/ipblock.go
```diff
@@ -232,22 +240,72 @@ func RunIpBlockUpdate(c *core.CommandConfig) error {
...
+	if allFlag {
+		err := DeleteAllIpBlocks(c)
+		if err != nil {
+			return err
+		}
+	} else {
+		if err := utils.AskForConfirm(c.Stdin, c.Printer, "delete ipblock"); err != nil {
+			return err
+		}
+		c.Printer.Verbose("Ip block with id: %v is deleting...", viper.GetString(core.GetFlagName(c.NS, cloudapiv5.ArgIpBlockId)))
```
AnaMariaMoga, line 254: For me, it makes more sense: "Deleting IpBlock with ID: %v..."

commands/cloudapi-v5/ipblock.go
gabisavu98: I think is a git misunderstanding, because I do not removed this line, is still there, and every resource has it,s return line that uses the Printer method

AnaMariaMoga: In my opinion, for the feedback to be given for both delete all and for delete simple - the `resp` variable needs to be returned as well on `DeleteAll<Resource>` function, and also to not be overwritten/defined as a new variable in the delete "simple" functionality.

commands/cloudapi-v5/backupunit.go
```diff
@@ -310,6 +327,47 @@ func getBackupUnitInfo(c *core.CommandConfig) *resources.BackupUnitProperties {
...
+		if err := utils.AskForConfirm(c.Stdin, c.Printer, "delete all the Backup Units"); err != nil {
+			return err
+		}
+		c.Printer.Verbose("Deleting all the BackupUnits...")
+		for _, backupUnit := range *backupUnitsItems {
+			if id, ok := backupUnit.GetIdOk(); ok && id != nil {
+				c.Printer.Verbose("Deleting Backup unit with id: %v...", *id)
+				resp, err = c.CloudApiV5Services.BackupUnit().Delete(*id)
+				if resp != nil {
+					c.Printer.Verbose(config.RequestTimeMessage, resp.RequestTime)
```
AnaMariaMoga, line 357: In the `delete-all` functionality, the idea is that multiple requests are being sent in the same command. Printing only the request time doesn't help that much the user to know where or what to debug. In the simple functionality, this is not  much a big of a problem, because at the end of the command, the request id is still printed. (with the getPrint() function on return). 

I suggest adding also a log for printing the Request Id (right before the log with the request time). It can be used maybe `GetId()` function from the printer or something similar.

This applies to all the places in this PR with the delete all functionality.

gabisavu98: For delete all, first you cans see the resources ids and names that will be deleted, and when is performing de delete operation, the id of the resource is shown again, so you can correlate between ids to see what resources were actually deleted.
I will add de request id too for easy debug purose

PR #121, opened 2021-10-22
Feat: implement --all flag for detach and remove commands

gabisavu98:
implemented --all flag for detach and remove commands

like --all flag for delete command, for every resource that has the detach or remove command, it will detach or remove all the resources associated to another resource or in data center case, all the data centers from a account

## Checklist

- [x] PR Title containing `Feat`/`Fix`/`Doc`
- [x] Tests added or updated
- [x] Documentation updated
- [x] SONAR Cloud Scan - warnings checked
- [x] Task updated (Jira/Github)

### DIFF
--- a/commands/cloudapi-v5/cdrom.go
+++ b/commands/cloudapi-v5/cdrom.go
@@ -147,7 +147,7 @@ Required values to run command:
 * Server Id
 * Cdrom Id`,
 		Example:    detachCdromServerExample,
-		PreCmdRun:  PreRunDcServerCdromIds,
+		PreCmdRun:  PreRunDcServerCdromDetach,
 		CmdRun:     RunServerCdromDetach,
 		InitClient: true,
 	})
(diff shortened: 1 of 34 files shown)


commands/cloudapi-v5/cdrom.go
```diff
@@ -224,27 +232,81 @@ func RunServerCdromGet(c *core.CommandConfig) error {
...
+					return nil, err
+				}
+				if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
+					return nil, err
+				}
+			}
+		}
 	}
-	return c.Printer.Print(getImagePrint(resp, c, nil))
+	return resp, err
```
AnaMariaMoga, line 309: If the error is already checked before, this is better to be `return resp, nil`

AnaMariaMoga (approved):

PR #145, opened 2021-12-29
fix: improve verbose print when delete/detach/remove --all

gabisavu98:

The print that is shown when you use a delete/detach/remove operation followed by the --all flag  shows now the id and the name of the resource which will be deleted in the same line

## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [ ] Sonar Cloud Scan
- [x] Changelog updated and version incremented (label: upcoming release)
- [x] Github Issue linked if any
- [ ] Jira task updated

### DIFF
--- a/commands/cloudapi-v6/nic.go
+++ b/commands/cloudapi-v6/nic.go
@@ -434,14 +434,16 @@ func DeleteAllNics(c *core.CommandConfig) (*resources.Response, error) {
 	}
 	if nicsItems, ok := nics.GetItemsOk(); ok && nicsItems != nil {
 		for _, nic := range *nicsItems {
+			toPrint := ""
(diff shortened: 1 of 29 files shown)


commands/cloudapi-v6/nic.go
```diff
@@ -434,14 +434,16 @@ func DeleteAllNics(c *core.CommandConfig) (*resources.Response, error) {
 	}
 	if nicsItems, ok := nics.GetItemsOk(); ok && nicsItems != nil {
 		for _, nic := range *nicsItems {
+			toPrint := ""
```
cristiGuranIonos, line 437: maybe this can be a function with the below - PrintNics

cristiGuranIonos (approved):

PR #201, opened 2022-09-23
feat: ftp image upload

avirtopeanu-ionos:

- [x] Add image upload command
- [x] Upload image to FTP server
     - [x] Find server URL dynamically via location flag
     - [x] Find server path to image file dynamically via image extension (iso to iso-images, all other to hdd-images)
     - [x] Upload small images
     - [x] Upload large images (Tested up to 100GB)
- [x] Upload images in parallel via goroutines
- [x] handle the ftp login more gracefully (see discussion in this PR https://github.com/ionos-cloud/ionosctl/pull/201#issuecomment-1273365799). Even if `IONOS_TOKEN` is set, Config file `userdata.username` and `userdata.password` fields should still be used as fallbacks for `IONOS_USERNAME` and `IONOS_PASSWORD`  (This bug doesn't affect any other resource since all commands so far supported bearer tokens)
- [x] Image Patch, Delete, Delete All
- [x] Update images after upload - use filtered list to find uploaded imgs UUIDs 
         Notes: contains technical debt. When support for multiple value per key filtering is added, this technical debt can be removed

## Checklist

- [ ] Tests added or updated
- [x] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


services/cloudapi-v6/resources/image.go
```diff
@@ -41,6 +64,47 @@ func NewImageService(client *Client, ctx context.Context) ImagesService {
...
+
+	// Do something with the FTP conn
+	err = c.Chdir(filepath.Dir(p.Path))
+	if err != nil {
+		fmt.Printf("Failed to change to %s\n", filepath.Dir(p.Path))
+		return err
+	}
+
+	data := bytes.NewBufferString("Hello World")
+	err = c.Upload(s.context, "test.iso", data)
```
ionos-landgraf-vin, line 98: s.context is nil that is the reason for the panic.

It would be good if we can default the name of the uploaded file to the basename of the file to upload, but it would be very handy if this uploaded name can be overwritten. This is especially true for CI pipelines, where always the same image name comes out of the pipeline but the upload name should be different all the time.

services/cloudapi-v6/resources/image.go
```diff
@@ -41,6 +64,47 @@ func NewImageService(client *Client, ctx context.Context) ImagesService {
...
+		Username:    viper.GetString(config.Username),
+		Passowrd:    viper.GetString(config.Password),
+		ExplicitTLS: true,
+		TLSConfig:   &tlsConfig,
+	}
+
+	//ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	//defer cancel()
+
+	c, err := ftps.Dial(context.TODO(), dialOptions)
```
ionos-landgraf-vin, line 84: you should use s.context here as well

commands/cloudapi-v6/image.go
```diff
@@ -104,9 +108,95 @@ func ImageCmd() *core.Command {
...
+	images := viper.GetStringSlice(core.GetFlagName(c.NS, "image"))
+	locations := viper.GetStringSlice(core.GetFlagName(c.NS, cloudapiv6.ArgLocation))
+	var eg errgroup.Group
+	for _, img := range images {
+		for _, loc := range locations {
+			url := fmt.Sprintf("ftp-%s.ionos.com", loc)
+			c.Printer.Verbose("Uploading %s to %s", img, url)
+
+			validHddImageExtensions := []string{".vmdk", ".vhd", ".vhdx", ".cow", ".qcow", ".qcow2", ".raw", ".vpc", ".vdi"}
+			// TODO: Move to prerun?
```
ionos-landgraf-vin, line 162: That would be good, first verify then upload

ionos-landgraf-vin (changes requested):

services/cloudapi-v6/resources/image.go
avirtopeanu-ionos: Ah, good find, thanks.

I was so focused on the context passed to "Dial" function that this other context slipped unnoticed

As for the auto basename, I had already been doing that (see prev. commits) but decided to temporarily always have the same image name for debugging purposes. However, the suggestion for having a customizable name via a flag is a great idea! 

Thanks!

ionos-landgraf-vin:
@avirtopeanu-ionos could "(Currently request times out due to large image size)" this be due to an inherited context timeout? maybe try a different context and see if there is a timeout then?

avirtopeanu-ionos:
@ionos-landgraf-vin USER seems to refer to login params. Currently, for logging in, viper looks up the following values: 
1. env variables `IONOS_USERNAME` & `IONOS_PASSWORD`,
2. config file `userdata.username`, `userdata.password` (attained by doing ionosctl login)

But I agree that is not really a great error message.

On top of adding a more specific error message, do you feel it is necessary to add flags for username and password logins (for this command alone)? I assume you're only using IONOS_TOKEN currently for all other commands

ionos-landgraf-vin:
These username and password are set in my environment via the config file, but I also have the `IONOS_TOKEN`, so in case only a token is present it doesn't seem to work. Code should be adopted to use username and password and also inform in case they are missing.

avirtopeanu-ionos:
Hi, yes, seems like the if `IONOS_TOKEN` is set, then config file username & password is ignored. This hasn't been a problem for any of the other commands since it was considered ok to unset your token to user username & password (which, as a side effect, would trigger reading from config file if `IONOS_USERNAME`, `IONOS_PASSWORD` unset)

gabisavu-ionos (approved):

commands/cloudapi-v6/image.go
```diff
@@ -159,7 +168,7 @@ func ImageCmd() *core.Command {
 		Aliases:    []string{"d"},
 		ShortDesc:  "Delete an image",
 		LongDesc:   "Use this command to delete a specified Image.\n\nRequired values to run command:\n\n* Image Id",
-		Example:    "placeholder", // TODO
+		Example:    "ionosctl image delete --image-id IMAGE_ID", // TODO
```
mimihalescu, line 171: should the TODO be removed? or is there anything to be added there?

mimihalescu (approved):

commands/cloudapi-v6/image.go
avirtopeanu-ionos: good catch, TODO should have been removed, thank you!

PR #205, opened 2022-10-03
Feat: image update, delete commands

avirtopeanu-ionos:

- Added image Update command, uses Go SDK Image Patch route
- Added Image Delete command
- Added Delete --all option 
- Minor improvements to SetFlag: description of flag will dynamically show which values user can choose from. (As a consequence, descriptions of flags should NOT contain a dot at the end ) 

## Checklist

- [ ] Tests added or updated
- [x] Documentation updated
- [x] Sonar Cloud Scan
- [ ] Jira task updated

### DIFF
--- a/commands/cloudapi-v6/image.go
+++ b/commands/cloudapi-v6/image.go
@@ -213,24 +213,27 @@ func RunImageDelete(c *core.CommandConfig) error {
 	}
 }
 
+// Util func - Given a slice of public & non-public images, return only those images that are non-public.
+// If any image in the slice has null properties, or "Properties.Public" field is nil, an error is thrown.
 func _getNonPublicImages(imgs *[]ionoscloud.Image) (*[]ionoscloud.Image, error) {
-	var publicImgs []ionoscloud.Image
+	var nonPublicImgs []ionoscloud.Image
 	for _, i := range *imgs {
 		properties, ok := i.GetPropertiesOk()
 		if !ok {
-			return nil, fmt.Errorf("failed to get properties of image\n")
+			return nil, fmt.Errorf("properties of image %s are nil\n", *i.GetId())
 		}
 		isPublic, ok := properties.GetPublicOk()
 		if !ok {
-			return nil, fmt.Errorf("failed to get public value of image %s\n", *i.GetId())
+			return nil, fmt.Errorf("public field of image %s is nil\n", *i.GetId())
 		}
 		if !*isPublic {
-			publicImgs = append(publicImgs, i)
+			nonPublicImgs = append(nonPublicImgs, i)
 		}
 	}
-	return &publicImgs, nil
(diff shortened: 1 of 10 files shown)


commands/cloudapi-v6/image.go
```diff
@@ -104,9 +109,274 @@ func ImageCmd() *core.Command {
...
+		if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
+			return err
+		}
+		return c.Printer.Print(getImagePrint(resp, c, nil))
+	}
+}
+
+// Util func - Given a slice of public & non-public images, return only those images that are non-public.
+// If any image in the slice has null properties, or "Properties.Public" field is nil, an error is thrown.
+func _getNonPublicImages(imgs *[]ionoscloud.Image) (*[]ionoscloud.Image, error) {
```
cristiGuranIonos, line 218: is this python naming?

commands/cloudapi-v6/image.go
```diff
@@ -104,9 +109,274 @@ func ImageCmd() *core.Command {
...
+		if id, ok := img.GetIdOk(); ok && id != nil {
+			c.Printer.Verbose("Starting deleting image with id: %v...", *id)
+			resp, err = c.CloudApiV6Services.Images().Delete(*id, resources.QueryParams{})
+			if resp != nil && printer.GetId(resp) != "" {
+				c.Printer.Verbose(config.RequestInfoMessage, printer.GetId(resp), resp.RequestTime)
+			}
+			if err != nil {
+				multiErr = multierr.Append(multiErr, fmt.Errorf(config.DeleteAllAppendErr, c.Resource, *id, err))
+				continue
+			} else {
```
cristiGuranIonos, line 288: doesn't seem you need an else for the print

commands/cloudapi-v6/image.go
```diff
@@ -104,9 +109,274 @@ func ImageCmd() *core.Command {
...
+			if err = utils.WaitForRequest(c, waiter.RequestInterrogator, printer.GetId(resp)); err != nil {
+				multiErr = multierr.Append(multiErr, fmt.Errorf(config.DeleteAllAppendErr, c.Resource, *id, err))
+				continue
+			}
+		}
+	}
+	if multiErr != nil {
+		return multiErr
+	}
+	return nil
```
cristiGuranIonos, line 300: just return multiErr

commands/cloudapi-v6/image.go
avirtopeanu-ionos: i know, right? This is the case for all resources. At first I wanted to change this, but kept it to be the same behaivour across all resources.

One more thing, I kept it the same as for other resources because when we eventually refactor this code, we can then do a find & replace with regex match.

PR #214, opened 2022-10-31
refactor: table headers remove tight coupling

avirtopeanu-ionos:
## What does this fix or implement?

improve code quality by reducing tight coupling when getting table headers

## Checklist

- [x] PR name added as appropriate (e.g. `feat:`/`fix:`/`doc:`/`test:`/`refactor:`)
- [x] Tests added or updated
- [x] Documentation updated
- [x] Sonar Cloud Scan
- [ ] Changelog updated and version incremented (label: upcoming release)
- [ ] Github Issue linked if any
- [x] Jira task updated

PR #215, opened 2022-11-03
Feat/certificate manager service

dgheorghioiu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [X] Jira task updated


commands/certmanager/patch.go
```diff
@@ -0,0 +1,70 @@
...
+	sdkgo "github.com/ionos-cloud/sdk-go-cert-manager"
+)
+
+var editProperties = sdkgo.CertificatePatchPropertiesDto{}
+
+func CertPatchCmd() *core.Command {
+	cmd := core.NewCommand(context.TODO(), nil, core.CommandBuilder{
+		Namespace:  "certmanager",
+		Resource:   "certificates",
+		Verb:       "patch",
```
avirtopeanu-ionos, line 18: can we please rename this to `update`, to keep in-line with the other commands?

commands/certmanager/post.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+
+	"github.com/ionos-cloud/ionosctl/pkg/core"
+	sdkgo "github.com/ionos-cloud/sdk-go-cert-manager"
+	"github.com/spf13/cobra"
+)
+
+var createProperties = sdkgo.CertificatePostPropertiesDto{}
+
+func CertPostCmd() *core.Command {
```
avirtopeanu-ionos, line 15: same as `patch` command: this should be `create`, but the actual user-side functionality seems fine

commands/certmanager/constants.go
```diff
@@ -0,0 +1,37 @@
+package certmanager
+
+// flags
+const (
+	CertId              = "certificate-id"
+	CertName            = "certificate-name"
```
avirtopeanu-ionos, line 6: We can shorten this to just `name`, which is a common flag (compute, k8s, to be used in mongo).
Move to pkg/constants (if this doesn't exist, rebase from master) 

Will be used as `constants.FlagName`

commands/certmanager/constants.go
```diff
@@ -0,0 +1,37 @@
...
+	CertChain           = "certificate-chain"
+	CertChainPath       = "certificate-chain-path"
+	PrivateKey          = "private-key"
+	PrivateKeyPath      = "private-key-path"
+	PostErrorFormatFlag = "%q requires at least 4 options by providing the required flag values or path values.\n\nUsage:\n%s\n%s\nFor more details, see '%s --help'."
+	PostErrorExample1   = "ionosctl certificate-manager create --certificate-name CERTIFICATE_NAME --certificate CERTIFICATE --certificate-chain CERTIFICATE_CHAIN --private-key PRIVATE_KEY"
+	PostErrorExample2   = "ionosctl certificate-manager create --certificate-name-path CERTIFICATE_NAME_PATH --certificate CERTIFICATE_PATH --certificate-chain-path CERTIFICATE_CHAIN_PATH --private-key-path PRIVATE_KEY_PATH\n"
+	AllFlag             = "all"
+	ArgCols             = "cols"
+	ArgWaitForState     = "wait-for-state"
```
avirtopeanu-ionos, line 19: Move to `constants` package, these already exist

commands/certmanager/constants.go
```diff
@@ -0,0 +1,37 @@
+package certmanager
+
+// flags
```
avirtopeanu-ionos, line 3: Flags should have `Flag` pre-pended to their variable name (or `flag` if unexported)

Like `FlagCertId`

commands/certmanager/constants.go
```diff
@@ -0,0 +1,37 @@
...
+	PrivateKeyPath      = "private-key-path"
+	PostErrorFormatFlag = "%q requires at least 4 options by providing the required flag values or path values.\n\nUsage:\n%s\n%s\nFor more details, see '%s --help'."
+	PostErrorExample1   = "ionosctl certificate-manager create --certificate-name CERTIFICATE_NAME --certificate CERTIFICATE --certificate-chain CERTIFICATE_CHAIN --private-key PRIVATE_KEY"
+	PostErrorExample2   = "ionosctl certificate-manager create --certificate-name-path CERTIFICATE_NAME_PATH --certificate CERTIFICATE_PATH --certificate-chain-path CERTIFICATE_CHAIN_PATH --private-key-path PRIVATE_KEY_PATH\n"
+	AllFlag             = "all"
+	ArgCols             = "cols"
+	ArgWaitForState     = "wait-for-state"
+)
+
+var RequiredFlagSets = [16][]string{{CertName, Cert, CertChain, PrivateKey},
```
avirtopeanu-ionos, line 22: Manually check these required flags in PreRun

commands/certmanager/get.go
```diff
@@ -0,0 +1,72 @@
...
+		Aliases:    []string{"g"},
+		ShortDesc:  "Get Certificate by ID",
+		LongDesc:   "Use this command to retrieve a Certificate by ID.",
+		Example:    "ionosctl certificate-manager get --certificate-id 47c5d9cc-b613-4b76-b0cc-dc531787a422",
+		PreCmdRun:  PreCmdGet,
+		CmdRun:     CmdGet,
+		InitClient: true,
+	})
+
+	cmd.AddStringFlag(CertId, "i", "", "Response get a single certificate (required)")
```
avirtopeanu-ionos, line 26: Don't hard-code `(required)` into the flag description. Instead, add `core.RequiredFlagOption()` as an argument to `AddStringFlag`:

`cmd.AddStringFlag(CertId, "i", "", "Response get a single certificate", core.RequiredFlagOption())`

Reason: Was hoping we could automate pre-runs someday, to just use these `core.Options` funcs, even though they are legacy code they seem to be a good idea, at least in theory. One could argue here YAGNI but there's only so much we can refactor at once.

commands/certmanager/post.go
```diff
@@ -0,0 +1,135 @@
...
+		RequiredFlagSets[10],
+		RequiredFlagSets[11],
+		RequiredFlagSets[12],
+		RequiredFlagSets[13],
+		RequiredFlagSets[14],
+		RequiredFlagSets[15],
+	)
+
+	if err != nil {
+		return fmt.Errorf(PostErrorFormatFlag,
```
avirtopeanu-ionos, line 127: This is reaaally hard to read. But once we replace the `core.CheckRequiredFlagsSets` above with manual verifs, I think we can have early returns which would replace this "handle-all' ErrorF string builder

pkg/utils/confirmation.go
```diff
@@ -40,6 +40,13 @@ func AskForConfirm(reader io.Reader, writer printer.PrintService, message string
...
+	//TODO: when responding no should not return error with invalid user input
+
+	//temporary solution
+	if answer == "n" || answer == "no" {
+		return errors.New("Stopped")
+	}
+
 	if answer != "y" && answer != "ye" && answer != "yes" {
 		return errors.New("invalid user input")
 	}
```
avirtopeanu-ionos, line 52: This is a pretty funny code snippet. I think we should scrap this altogether. But, let's handle this in a different PR. Could you please revert this? You can leave a TODO flag here, if you wish

commands/certmanager/post.go
```diff
@@ -0,0 +1,135 @@
...
+		Aliases:    []string{"c"},
+		ShortDesc:  "Add a new Certificate",
+		LongDesc:   "Use this command to add a Certificate.",
+		Example:    "ionosctl certificate-manager create",
+		PreCmdRun:  PreCmdPost,
+		CmdRun:     CmdPost,
+		InitClient: true,
+	})
+
+	cmd.AddStringFlag(CertName, "", "", "Specify name of the certificate (required either this or --certificate-name-path)")
```
avirtopeanu-ionos, line 29: Could you please add some shorthand flags? They don't need really need to be defined as constants, but it would help to do so, since we usually want to keep track of which shorthand flag is bound to which flag

commands/certmanager/certificates.go
```diff
@@ -0,0 +1,156 @@
...
+func printProperties(value *ionoscloud.CertificateDto, c *core.CommandConfig, flags []bool) string {
+	var printString string
+	if flags[0] {
+		printString += *value.Properties.Certificate
+	}
+	if flags[1] {
+		printString += *value.Properties.CertificateChain
+	}
+	return printString
+}
```
avirtopeanu-ionos, line 59: 1. Don't concatenate these into a single string. return either one or the other (If both are set, it's ok, print one of them, we can mark them as exclusive flags in the commands)
2. instead of taking a `[]bool` it would make the code easier to understand if we take two bools, `isSetPrintCert`, `isSetPrintChain` or whatever better naming comes to mind
3. Is this func used anywhere other than in the `get` cmd? if not, we can just move the functionality there and no need to use a func. Otherwise, fi it's used in multiple places, a func is nice, it helps reduce code duplication. Nice!

commands/certmanager/post.go
```diff
@@ -0,0 +1,135 @@
...
+		return propertyValue, nil
+	}
+	return propertyValue, nil
+}
+
+func CmdPost(c *core.CommandConfig) error {
+	c.Printer.Verbose("Adding Certificate...")
+	var name, certificate, certificateChain, privateKey string
+
+	name, err := SetProperties(c, CertName)
```
avirtopeanu-ionos, line 71: `SetProperties` functionality is nice. However, it's a bit unclear. We are lazy devs and want to know what a function does just by its signature. Also, the fact that it looks into `{certName}-path` too is not "advertised".

1. Rename to reflect above. What do you think about `GetPropertyWithFallback`?
2. Hardcoding the fact that `{certName}-path` is a fallback flag isn't good. Take the path params as an argument too: `GetPropertyWithFallback(c, certName, certNamePath)`

commands/certmanager/post.go
```diff
@@ -0,0 +1,135 @@
...
+		RequiredFlagSets[7],
+		RequiredFlagSets[8],
+		RequiredFlagSets[9],
+		RequiredFlagSets[10],
+		RequiredFlagSets[11],
+		RequiredFlagSets[12],
+		RequiredFlagSets[13],
+		RequiredFlagSets[14],
+		RequiredFlagSets[15],
+	)
```
avirtopeanu-ionos, line 124:
```
if !viper.IsSet(flagCertificate) && !viper.IsSet(flagCertificatePath) { 
    return fmt.Errorf("Man, give me a certificate!!!!!!")
 }
```

Repeat above for each flag you want to add path support to. Bonus points if you make a function to do these checks but might not be necessary for just ~3 flags.

By the way, do you also provide file reading support for `FlagName`? This isn't really needed, certificate names are supposed to be human readable anyways. They'll be at most a few dozen characters long, no need to read from file for that.

avirtopeanu-ionos (changes requested):

commands/certmanager/apiversion.go
```diff
@@ -0,0 +1,39 @@
...
+		Example:    "ionosctl certificate-manager api-version",
+		PreCmdRun:  core.NoPreRun,
+		CmdRun:     CmdGetApiVersion,
+		InitClient: true,
+	})
+
+	_ = cmd.Command.RegisterFlagCompletionFunc(ArgCols, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return allCols, cobra.ShellCompDirectiveNoFileComp
+	})
+
```
avirtopeanu-ionos, line 28: add ArgNoHeaders, maybe I (an IonosCTL user) want to do a curl on the swagger / openapi-spec: I can't because I can't get rid of table headers

commands/certmanager/certificates.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		o := structs.Map(certPrint)
+		out = append(out, o)
+	}
+	return out
+}
+
+var allCols = structs.Names(CertPrint{})
+
+func getCertHeaders(customColumns []string) []string {
```
avirtopeanu-ionos, line 101: You can rebase (or cherry-pick, or merge) https://github.com/ionos-cloud/ionosctl/pull/214 and get rid of this function, and in `getCertPrint` instead of this func you can use:

`printer.GetHeadersAllDefault(allCols, viper.GetStringSlice(core.GetGlobalFlagName(c.Resource, constants.ArgCols)))`

commands/certmanager/certificates.go
```diff
@@ -0,0 +1,156 @@
...
+		if versionOk, ok := api.GetVersionOk(); ok && versionOk != nil {
+			apiPrint.Version = *versionOk
+		}
+		o := structs.Map(apiPrint)
+		out = append(out, o)
+	}
+	return out
+}
+
+var allAPICols = structs.Names(ApiPrint{})
```
avirtopeanu-ionos, line 148: I vote that we should move any API-print related things to `apiversion.go`, they pollute this file when they're really only needed in that specific `apiversion.go` file

avirtopeanu-ionos (changes requested):

commands/certmanager/certificates.go
dgheorghioiu-ionos: This function prints only one option.

commands/certmanager/constants.go
dgheorghioiu-ionos: This is already used only in PreRun

commands/certmanager/post.go
dgheorghioiu-ionos: I thought I should keep the same error format as other commands.

services/certmanager/resources/certificates.go
```diff
@@ -0,0 +1,71 @@
...
+	return &certsService{
+		client:  client,
+		context: ctx,
+	}
+}
+
+func (svc *certsService) Get(certId string) (sdkgo.CertificateDto, *sdkgo.APIResponse, error) {
+	req := svc.client.CertificatesApi.CertificatesGetById(svc.context, certId)
+	cert, res, err := svc.client.CertificatesApi.CertificatesGetByIdExecute(req)
+	return cert, res, err
```
cristiGuranIonos, line 40: This function can be one line I think : return svc.client.CertificatesApi.CertificatesGetById(svc.context, certId).Execute()

commands/certmanager/post.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return cmd
+}
+
+func GetPropertyWithFallback(c *core.CommandConfig, property string, propertyPath string) (string, error) {
+	propertyValue, err := c.Command.Command.Flags().GetString(property)
+	if err != nil {
+		return "", err
+	}
+	if propertyValue == "" {
```
avirtopeanu-ionos, line 47: Here I would prefer early returns for readability, but its up to you:

```
if propertyValue != "" {
    return propertyValue, nil
}
// propertyValuePath handling below
```

https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

commands/certmanager/post.go
```diff
@@ -0,0 +1,123 @@
...
+	cert, _, err := c.CertificateManagerServices.Certs().Post(*Dto)
+	if err != nil {
+		return err
+	}
+
+	return c.Printer.Print(getCertPrint(nil, c, &[]sdkgo.CertificateDto{cert}))
+}
+
+// couldn't think of a better solution to this
+func PreCmdPost(c *core.PreCommandConfig) error {
```
avirtopeanu-ionos, line 100: Nice!! Way more readable than the matrix solution :+1: :+1:

commands/certmanager/post.go
```diff
@@ -0,0 +1,123 @@
...
+		err = fmt.Errorf("%veither --%s or --%s must be set\n", err, FlagCert, FlagCertPath)
+	}
+	if !viper.IsSet(core.GetFlagName(c.NS, FlagCertChain)) && !viper.IsSet(core.GetFlagName(c.NS, FlagCertChainPath)) {
+		err = fmt.Errorf("%veither --%s or --%s must be set\n", err, FlagCertChain, FlagCertChainPath)
+	}
+	if !viper.IsSet(core.GetFlagName(c.NS, FlagPrivateKey)) && !viper.IsSet(core.GetFlagName(c.NS, FlagPrivateKeyPath)) {
+		err = fmt.Errorf("%veither --%s or --%s must be set\n", err, FlagPrivateKey, FlagPrivateKeyPath)
+	}
+	if !viper.IsSet(core.GetFlagName(c.NS, FlagCertName)) {
+		err = fmt.Errorf("%veither --%s must be set\n", err, FlagCertName)
```
avirtopeanu-ionos, line 112: `FlagCertName` can just use cobra's markFlagAsRequired func, since it doesn't have a Path fallback

commands/certmanager/post.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	privateKey, err = GetPropertyWithFallback(c, FlagPrivateKey, FlagPrivateKeyPath)
+	if err != nil {
+		return err
+	}
+
+	createProperties.SetName(name)
+	createProperties.SetCertificate(certificate)
+	createProperties.SetCertificateChain(certificateChain)
+	createProperties.SetPrivateKey(privateKey)
```
avirtopeanu-ionos, line 85: Nice! :1st_place_medal:

services/certmanager/resources/client.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type clientService struct {
+	client *sdkgo.APIClient
+}
+
+var _ ClientService = &clientService{}
+
+func NewClientService(name, pwd, token, hostUrl string) (ClientService, error) {
+	if token == "" && (name == "" || pwd == "") {
```
cristiGuranIonos, line 33: I think if the err check is moved outside the init chain, you wouldn't need to throw an error inside the constructor. To remember as a refactor at some point. Also the err message can be improved to mention that you need to provide either username and pass, or token.

commands/certmanager/certificates.go
```diff
@@ -0,0 +1,98 @@
...
+func printProperties(value *ionoscloud.CertificateDto, c *core.CommandConfig, flags []bool) string {
+	var printString string
+	if flags[0] {
+		printString += *value.Properties.Certificate
+	}
+	if flags[1] {
+		printString += *value.Properties.CertificateChain
+	}
+	return printString
+}
```
avirtopeanu-ionos, line 58: - No bool slice - It allows passing an infinite number of bools. Must restrict to "either this or that flag"

- Dont concatenate into printString. If I pass printProperties(true, true) it would return both cert and cert chain.

- Use the SDK getters ` *value.GetProperties().getCertificateChain()`

- Param `c` not used - remove

- Signature should match behaivour:
    -  Func doesn't print anything: prefix `get` instead. 
    - require only one bool param instead of a slice of unknown length. If this bool set, print Cert. If not, print CertChain

```
func getOnlyCertOrCertChain(value *ionoscloud.CertificateDto, getCertElseChain bool) string {
	if getCertElseChain {
		return *value.getProperties().getCertificate()
	}
	return *value.getProperties().getCertificateChain()
}
```

avirtopeanu-ionos (changes requested):

services/certmanager/resources/certificates.go
avirtopeanu-ionos: @cristiGuranIonos I agree with this but this is the case for every single service in IonosCTL

To be frank, I don't see a good reason for having services, seeing as the SDKs we import basically act as a service layer anyway

The only thing these services manage to accomplish is generate more "paperwork" for the developer IMHO

Seeing as this problem permeates into the entirety of the codebase, what do you think of handling in a different PR @cristiGuranIonos ?

services/certmanager/resources/client.go
avirtopeanu-ionos: @cristiGuranIonos again, fully agree with this.... buuuut, this is the case for every single service. Same discussion as above

Duplicated code:

https://github.com/ionos-cloud/ionosctl/blob/master/services/cloudapi-v6/resources/client.go#L34-L45

https://github.com/ionos-cloud/ionosctl/blob/master/services/dbaas-postgres/resources/client.go#L32-L41

Again, maybe handle this in a different PR for consistency sake?

services/certmanager/resources/certificates.go
cristiGuranIonos: yes, definitely. As part of a refactor.

services/certmanager/resources/client.go
cristiGuranIonos: I saw, yes definitely in another PR :)

commands/certmanager/delete.go
```diff
@@ -0,0 +1,69 @@
...
+	_ = cmd.Command.RegisterFlagCompletionFunc(config.ArgCols, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return allCols, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	return cmd
+}
+
+func CmdDelete(c *core.CommandConfig) error {
+	var err error
+	allFlag := c.Command.Command.Flag("all").Changed
```
avirtopeanu-ionos: `.Changed` -- looking at pflag documentation: https://pkg.go.dev/github.com/spf13/pflag#Flag.Changed - `// If the user set the value (or if left to default)` 

would this surely not trigger "delete all" even if we want to delete a single certificate?

Why not use `cmd.Command.Flags().GetBool(constants.ArgAll)`?

avirtopeanu-ionos: One more thing to note here, if using `.Changed` - using `ionosctl cert-manager delete --all=false` would probably also trigger `delete all`

commands/certmanager/delete.go
```diff
@@ -0,0 +1,69 @@
...
+
+func CertDeleteCmd() *core.Command {
+	cmd := core.NewCommand(context.TODO(), nil, core.CommandBuilder{
+		Namespace: "certmanager",
+		Resource:  "certificates",
+		Verb:      "delete",
+		Aliases:   []string{"g"},
+		ShortDesc: "Delete Certificate by ID",
+		LongDesc:  "Use this command to delete a Certificate by ID.",
+		Example:   "ionsoclt certificate-manager delete --certificate-id 12345678-1234-1234-1234-123456789012",
```
avirtopeanu-ionos: misspelled `ionosctl`

commands/certmanager/apiversion.go
```diff
@@ -0,0 +1,99 @@
...
+		Aliases:    []string{"api", "info"},
+		ShortDesc:  "Get Certificate Manager API Version",
+		LongDesc:   "Use this command to retrieve API Version.",
+		Example:    "ionosctl certificate-manager api-version",
+		PreCmdRun:  core.NoPreRun,
+		CmdRun:     CmdGetApiVersion,
+		InitClient: true,
+	})
+
+	cmd.AddBoolFlag(constants.ArgNoHeaders, "n", false, "Response delete all certificates")
```
avirtopeanu-ionos, line 27: `Response delete all certificates` -> NoHeaders description. Can use a constant

commands/certmanager/delete.go
```diff
@@ -0,0 +1,76 @@
...
+		Aliases:    []string{"d"},
+		ShortDesc:  "Delete Certificate by ID or all Certificates",
+		LongDesc:   "Use this command to delete a Certificate by ID.",
+		Example:    "ionsoctl certificate-manager delete --certificate-id 47c5d9cc-b613-4b76-b0cc-dc531787a422",
+		PreCmdRun:  PreCmdDelete,
+		CmdRun:     CmdDelete,
+		InitClient: true,
+	})
+
+	cmd.AddStringFlag(FlagCertId, constants.FlagIdP, "", "Response delete a single certificate (required)")
```
avirtopeanu-ionos, line 26: `cmd.AddStringFlag(FlagCertId, constants.FlagIdP, "", "Certificate ID", core.RequiredFlagOption)`

avirtopeanu-ionos (approved):
Great work!! :partying_face:

PR #216, opened 2022-11-04
fix: config file non-fatal edge cases, tests

avirtopeanu-ionos:

certain edge cases for reading config file would fatally stop the execution when instead should recover

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -12,19 +12,172 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestUsingJustTokenEnvVar(t *testing.T) {
+	os.Clearenv()
+	viper.Reset()
+
+	viper.SetConfigFile(os.DevNull)
+	viper.Set(constants.ArgConfig, os.DevNull)
+
+	assert.NoError(t, os.Setenv(sdk.IonosTokenEnvVar, "tok"))
+	assert.NoError(t, Load())
+	assert.Equal(t, "tok", viper.GetString(Token))
+	assert.Equal(t, "", viper.GetString(Username))
+	assert.Equal(t, "", viper.GetString(Password))
+	assert.Equal(t, "", viper.GetString(ServerUrl))
+}
+
+func TestTokEnvWithUserPassConfigBackup(t *testing.T) {
+	// Useful for API routes which don't accept bearer tokens, or custom IonosCTL commands (Image Upload)
+	os.Clearenv()
+	viper.Reset()
+
+	assert.NoError(t, os.Setenv(sdk.IonosTokenEnvVar, "tok"))
+	path := filepath.Join("..", "testdata", "config_user_pass.json") // TODO: These files should be created and then destroyed by the tests
+	viper.SetConfigFile(path)
+	viper.Set(constants.ArgConfig, path)
+	assert.NoError(t, os.Chmod(path, 0600))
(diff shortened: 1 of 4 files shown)


gabisavu-ionos (approved):

pkg/config/config_test.go
```diff
@@ -12,19 +12,156 @@ import (
...
+func TestTokEnvWithFullConfig(t *testing.T) {
+	// Config token should not override env var token
+	os.Clearenv()
+	viper.Reset()
+
+	assert.NoError(t, os.Setenv(sdk.IonosTokenEnvVar, "tok"))
+	path := filepath.Join("..", "testdata", "config.json") // TODO: These files should be created and then destroyed by the tests
+	viper.SetConfigFile(path)
+	viper.Set(constants.ArgConfig, path)
+	assert.NoError(t, os.Chmod(path, 0600))
```
adeatcu-ionos, line 57: @avirtopeanu-ionos not a big deal, you can leave it as it is, but since now you are only reading from the file, the number can be `0400`. You can leave it as it is since in the future you will create the test file, write some info in it, use the test file for the test and then delete the test file.

pkg/config/config_test.go
```diff
@@ -12,19 +12,156 @@ import (
...
+	assert.Equal(t, "", viper.GetString(ServerUrl))
+}
+
+func TestTokEnvWithUserPassConfigBackup(t *testing.T) {
+	// Useful for API routes which don't accept bearer tokens, or custom IonosCTL commands (Image Upload)
+	os.Clearenv()
+	viper.Reset()
+
+	assert.NoError(t, os.Setenv(sdk.IonosTokenEnvVar, "tok"))
+	path := filepath.Join("..", "testdata", "config_user_pass.json") // TODO: These files should be created and then destroyed by the tests
```
adeatcu-ionos, line 36: @avirtopeanu-ionos good mention for `TODO`.

pkg/config/config_test.go
avirtopeanu-ionos: there had been a discussion previouslyabout the config file permissions. The idea is that LoadFile() fails if the perms aren't exactly `0600`. I can't comment on why this is exactly, but this is the reason why we set exactly this perm number

pkg/config/config_test.go
```diff
@@ -12,19 +12,156 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestUsingJustTokenEnvVar(t *testing.T) {
+	os.Clearenv()
```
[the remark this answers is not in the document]
avirtopeanu-ionos: good call. On my end these foobar env vars dont carry over. Checking official go tests there doesn't seem to be any special "cleaning procedure" in place https://github.com/golang/go/blob/6fd0520db3cffad4fd01c98319b9b1bf3c50f5e2/src/cmd/go/go_test.go#L280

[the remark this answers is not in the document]
avirtopeanu-ionos: https://unix.stackexchange.com/questions/372627/is-it-possible-to-pass-environment-variables-from-child-to-parent-in-user-space

mimihalescu (approved):

avorima:
Can you make a release ASAP?

PR #227, opened 2023-01-09
added MaxResults for Group Shares

dgheorghioiu-ionos:
## Checklist

- [X] PR name added as appropriate (e.g. `feat:`/`fix:`/`doc:`/`test:`/`refactor:`)
- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Changelog updated and version incremented (label: upcoming release)
- [ ] Github Issue linked if any
- [X] Jira task updated

PR #229, opened 2023-01-10
fix: listing of group resources

webner:

Fixes the issue that listing of group resources always returned an empty list. (max-results defaulted to 0)

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


avirtopeanu-ionos (approved):
Thanks for this PR! Similar to #227. Perhaps in the future we can refactor this flag so that MaxResults is added programmatically to all list commands, rather than duplicated across all resources.

PR #230, opened 2023-01-10
fix: write text warnings to stderr

webner:

Fixes the issue that text warnings or send to stdout which makes it hard to parse the output in a script.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -65,17 +65,17 @@ type JSONPrinter struct {
 	Stderr io.Writer
 }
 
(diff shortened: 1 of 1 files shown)


avirtopeanu-ionos (approved):
I think adding this behavior for `-o text` is a great idea (before, this was only the case for `-o json`). Also, I like the `write` utility function, it reduces code duplication. Thanks for this PR!

PR #235, opened 2023-01-26
Refactor/client object

dgheorghioiu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [X] Jira task updated


pkg/config/config.go
```diff
@@ -145,3 +150,49 @@ func configFileWriter() (io.WriteCloser, error) {
...
+const depthQueryParam = int32(5)
+
+type Client struct {
+	CloudClient       *cloudv6.APIClient
+	AuthClient        *sdkgoauth.APIClient
+	CertManagerClient *certmanager.APIClient
+	DbaasClient       *dbaas.APIClient
+}
+
+func NewClient(name, pwd, token, hostUrl string) (*Client, error) {
```
avirtopeanu-ionos, line 163: Ideally, this function shouldn't be visible outside of this package - it should be illegal to be able to bypass the `GetClient` func (which is the one making sure there is never more than one instance of a Client). That being said though, I am aware the tests are very resistant to change, so if you think un-exporting `NewClient` is too much effort (because of the implied effort in changing the tests to use the `GetClient` func), we can do this part in some later PR when the tests become easier to manage

avirtopeanu-ionos: As a compromise, I think we can add a comment to prevent future devs from ever using this function in new code, and to point them toward `GetClient` instead

pkg/config/config.go
```diff
@@ -145,3 +150,49 @@ func configFileWriter() (io.WriteCloser, error) {
...
+	var err error
+	once.Do(func() {
+		err = Load()
+		instance, err = NewClient(viper.GetString(Username), viper.GetString(Password), viper.GetString(Token), GetServerUrl())
+	})
+	if err != nil {
+		return nil, err
+	}
+	return instance, nil
+}
```
avirtopeanu-ionos, line 198: very nice! 

To make sure we're good in the future too, could you please add some test cases for this func? We need to test that empty credentials throws an understandable error, and we also need to check that a single Client is created even though GetClient is called however many times.

commands/certmanager/certmanager_test.go
```diff
@@ -130,19 +129,13 @@ func TestCertificateManagerServiceCmd(t *testing.T) {
...
-			certs, _, err := svc.Get().CertificatesApi.CertificatesGet(context.Background()).
-				Execute()
+			svc, err := config.GetClient()
+			certs, _, err := svc.CertManagerClient.CertificatesApi.CertificatesGet(context.Background()).Execute()
 			assert.NoError(t, err)
 
 			var id string
 			for _, dto := range *certs.GetItems() {
-				if *dto.GetProperties().GetName() == "certificate" {
+				if *dto.GetProperties().GetName() == "___test___certificate___test___" {
```
cristiGuranIonos, line 138: what does this do?

pkg/config/config.go
```diff
@@ -145,3 +150,73 @@ func configFileWriter() (io.WriteCloser, error) {
...
+		CertManagerClient: certmanager.NewAPIClient(certManagerConfig), DbaasClient: dbaas.NewAPIClient(dbaasConfig)}, nil
+}
+
+var once sync.Once
+var instance *Client
+
+func GetClient() (*Client, error) {
+	var err error
+	once.Do(func() {
+		err = Load()
```
cristiGuranIonos, line 191: isn't err lost here?

pkg/config/config.go
```diff
@@ -145,3 +150,73 @@ func configFileWriter() (io.WriteCloser, error) {
...
+
+	return &Client{CloudClient: cloudv6.NewAPIClient(clientConfig), AuthClient: sdkgoauth.NewAPIClient(authConfig),
+		CertManagerClient: certmanager.NewAPIClient(certManagerConfig), DbaasClient: dbaas.NewAPIClient(dbaasConfig)}, nil
+}
+
+var once sync.Once
+var instance *Client
+
+func GetClient() (*Client, error) {
+	var err error
```
cristiGuranIonos, line 189: doesn't this mean err is thown only on first call of the function and nil on the others?

PR #252, opened 2023-03-13
feat: container registry

dgheorghioiu-ionos:
## Checklist

- [X] Tests added or updated
- [X] Documentation updated
- [x] Sonar Cloud Scan
- [ ] Jira task updated


tools/container-registry/doc.go
```diff
@@ -0,0 +1,212 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ionos-cloud/ionosctl/v6/commands"
+	"github.com/ionos-cloud/ionosctl/v6/pkg/core"
+)
+
+func main() {
```
avirtopeanu-ionos, line 15: There were some changes to master which allow you to reduce the code in this `doc.go` file

https://github.com/ionos-cloud/ionosctl/blob/4574e05b81b61b5e49eef97dcc2865ddd3dbcc59/tools/dataplatform/doc.go

commands/container-registry/registry/post.go
```diff
@@ -0,0 +1,111 @@
...
+	)
+	_ = cmd.Command.RegisterFlagCompletionFunc(
+		FlagRegGCDays,
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return []string{
+				"Modnday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday",
+			}, cobra.ShellCompDirectiveNoFileComp
+		},
+	)
+	cmd.AddStringFlag(FlagRegGCTime, "", "", "Specify the garbage collection schedule time of day")
```
avirtopeanu-ionos, line 52: would be nice to tell the users what time format we expect. afaik it is 00:00:00? could even add a few completer options in that format, but that's up to you

commands/container-registry/container-registry_test.go
```diff
@@ -0,0 +1,81 @@
...
+		err = errors.New("non-available cmd")
+	}
+	assert.NoError(t, err)
+	if ok := location.RegLocationsListCmd().IsAvailableCommand(); !ok {
+		err = errors.New("non-available cmd")
+	}
+	assert.NoError(t, err)
+	if ok := repository.RegRepoDeleteCmd().IsAvailableCommand(); !ok {
+		err = errors.New("non-available cmd")
+	}
```
avirtopeanu-ionos, line 80: Why not do `assert.True(t, repository.RegRepoDeleteCmd().IsAvailableCommand())` ?

commands/container-registry/registry/registry_test.go
```diff
@@ -0,0 +1,84 @@
...
+		registries, _, err := svc.RegistryClient.RegistriesApi.RegistriesGet(context.Background()).Execute()
+		assert.NoError(t, err)
+
+		var newRegistry *ionoscloud.RegistryResponse
+		for _, registry := range *registries.GetItems() {
+			if *registry.GetProperties().GetName() == name {
+				newRegistry = &registry
+			}
+		}
+
```
avirtopeanu-ionos, line 41: I think it would be great if we could add an assert that either: newRegistry is not nil (i.e. the if-block was entered least once), or that newRegistry's name is `name` - this second option would cause a panic though, if the if-block isn't entered at at all

avirtopeanu-ionos: sorry. my bad, we're using `newRegistry` below. this works fine. Nice!

commands/container-registry/name/names.go
```diff
@@ -0,0 +1,50 @@
...
+	name, err := c.Command.Command.Flags().GetString("name")
+	if err != nil {
+		return err
+	}
+	res, _ := c.ContainerRegistryServices.Name().Head(name)
+	if res.StatusCode == 404 {
+		return c.Printer.Print("Name is available.")
+	}
+	if res.StatusCode == 400 {
+		return c.Printer.Print("Bad request.")
```
avirtopeanu-ionos, line 38: what exactly is wrong with the request? perhaps print the error in this case, or some other API message to give a hint of what's wrong

```
 % ionosctl cr names --name hi
Bad request.
```

commands/container-registry/repository/repository.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/ionos-cloud/ionosctl/v6/pkg/core"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func RegRepoDeleteCmd() *core.Command {
+	cmd := core.NewCommand(
+		context.TODO(), nil, core.CommandBuilder{
+			Namespace: "container-registry",
+			Resource:  "repository",
```
avirtopeanu-ionos, line 15: it's not clear from the command name that this command is a DELETE operation. Yes, the aliases and the description point out that in fact that is deleting the repo contents, but for that you need to actually type out the command.

As a solution, either rename to `repository-delete`, or maybe add a subcommand: `repository delete`? 

Note;   If we use `repository-delete`. we would have to make a breaking change if some other subcommand is added to `repo` namespace

commands/container-registry/registry/post.go
```diff
@@ -0,0 +1,111 @@
...
+
+	cmd.Command.Flags().StringSlice(constants.ArgCols, nil, printer.ColsMessage(allCols))
+	_ = cmd.Command.RegisterFlagCompletionFunc(
+		constants.ArgCols,
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return allCols, cobra.ShellCompDirectiveNoFileComp
+		},
+	)
+	cmd.AddStringFlag(FlagName, "n", "", "Specify the name of the registry", core.RequiredFlagOption())
+	cmd.AddStringFlag(FlagLocation, "", "", "Specify the location of the registry", core.RequiredFlagOption())
```
avirtopeanu-ionos, line 39: completion would be nice, in case more locations are added in the future

commands/container-registry/registry/list.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func RegListCmd() *core.Command {
+	cmd := core.NewCommand(
+		context.TODO(), nil, core.CommandBuilder{
+			Namespace:  "container-registry",
+			Resource:   "registry",
+			Verb:       "list",
```
avirtopeanu-ionos, line 17: can't filter over the results

commands/container-registry/token/tokens.go
```diff
@@ -0,0 +1,184 @@
...
+	return r
+}
+
+type TokenPrint struct {
+	TokenId             string `json:"TokenId,omitempty"`
+	DisplayName         string `json:"DisplayName,omitempty"`
+	ExpiryDate          string `json:"ExpiryDate,omitempty"`
+	CredentialsUsername string `json:"CredentialsUsername,omitempty"`
+	CredentialsPassword string `json:"CredentialsPassword,omitempty"`
+	Status              string `json:"Status,omitempty"`
```
avirtopeanu-ionos, line 93: would be nice to have a "scopes" column - but if it's not feasible to unmarshal it all into a column then it's ok to skip it

commands/container-registry/token/tokens.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		o := structs.Map(tokenPrint)
+		out = append(out, o)
+	}
+	return out
+}
+
+var allCols = structs.Names(TokenPrint{})
+
+// TODO: check how other resources are doing this
```
avirtopeanu-ionos, line 130: Other resources take the given parent resource as a param for the completer, and use that to make a GET for the child object. In this case, you'd take the registry ID as a param.

However, this is also fine

commands/container-registry/token/tokens.go
```diff
@@ -0,0 +1,184 @@
...
+		} else if expiryTime[i] == 'd' {
+			days, _ = strconv.Atoi(number)
+			number = ""
+		} else if expiryTime[i] == 'h' {
+			hours, _ = strconv.Atoi(number)
+			number = ""
+		}
+	}
+
+	return time.Duration(years)*time.Hour*24*365 + time.Duration(months)*time.Hour*24*30 + time.Duration(days)*time.Hour*24 + time.Duration(hours)*time.Hour, nil
```
avirtopeanu-ionos, line 183: why not parse the string using time pkg ? into a duration

dgheorghioiu-ionos: Valid time units for time.ParseDuration are "ns", "us" (or "µs"), "ms", "s", "m" or "h".
I have to parse manually larger time units.

avirtopeanu-ionos (approved):

PR #274, opened 2023-04-03
feat: vcpu

dgheorghioiu-ionos:
## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [X] Jira task updated


mimihalescu (approved):

commands/cloudapi-v6/server.go
```diff
@@ -176,9 +192,9 @@ You can wait for the Request to be executed using ` + "`" + `--wait-for-request`
 	_ = create.Command.RegisterFlagCompletionFunc(cloudapiv6.ArgTemplateId, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		return completer.TemplatesIds(os.Stderr), cobra.ShellCompDirectiveNoFileComp
 	})
-	create.AddStringFlag(cloudapiv6.ArgType, "", serverEnterpriseType, "Type usages for the Server")
+	create.AddStringFlag(cloudapiv6.ArgType, "", serverEnterpriseType, "Type usages for the Server. Can be one of: ENTERPRISE, CUBE or VCPU")
```
avirtopeanu-ionos, line 195: You can (but aren't forced to) use our custom `AddSetFlag` to limit to certain values and automatically add completions and "Can be one of..." help text for those values

commands/cloudapi-v6/server.go
```diff
@@ -514,6 +530,9 @@ func PreRunServerCreate(c *core.PreCommandConfig) error {
 		return err
 	}
 	// Validate flags
+	if viper.GetString(core.GetFlagName(c.NS, cloudapiv6.ArgType)) == serverVCPUType && viper.IsSet(core.GetFlagName(c.NS, constants.FlagCpuFamily)) {
```
avirtopeanu-ionos, line 533: sooo, regarding setting / not setting `--cpu-family` here

the current approach here is fine -- buuut, it is a bit inconsistent with the `CUBE` handling - where the default is that the flag is unset and the user is free to pick a value. 

For the CUBE type handling, the context is that CPU_FAMILY was meant to always be set as `INTEL_SKYLAKE`  (in the past) - but a breaking change occured in the API - so the CUBE `--cpu-family` behaviour is meant as a way to not force us to come up with a hotfix on the ionosctl side.

Could you either implement the same behaviour as for CUBE servers here (where the user can still freely set `--cpu-family` - but the default is that it is an unset flag) - for forward compatibility purposes? Either that, or throw a similar error for CUBE servers too

avirtopeanu-ionos (approved):

PR #287, opened 2023-05-04
feat: DNS API support

avirtopeanu-ionos:
DNS API support


pkg/config/config.go
```diff
@@ -26,21 +26,28 @@ func GetUserData() map[string]string {
...
 
-// GetServerUrl returns the API URL from flags, config or env in order of priority.
-// The caller must ensure to load config or env vars beforehand, so they can be included.
+// GetServerUrl returns the server URL the SDK should use, with support for layered fallbacks.
 //
-// Priority:
-// 1. Explicit flag
-// 2. Env/config file
-// 3. Flag default value
+// ⚠️ WARNING: NEVER use viper.Get/os.Getenv to try to bypass this function! ⚠️
```
cristiGuranIonos, line 31: maybe add an explanation as to what happens when you bypass

avirtopeanu-ionos: there is a note, about one of the behaviours: 

> // In older versions this env var was bound to Viper. Viper does not support case-sensitive keys.
// This means someone could export `IoNoS_API_url` and it would work. So - don't try it. Would be breaking.

also, it could be that the variable is empty, when instead we want to fallback to the next config layer, so I guess I'll also add a note for that

this func is cherry-picked from #295 , since DNS will be released before the config file changes, and the DNS API was being overriden to be `api.ionos.com` because of this func.

avirtopeanu-ionos: there we go, added notes:

```
// GetServerUrl returns the server URL the SDK should use, with support for layered fallbacks.
//
// ⚠️ WARNING: NEVER use viper.Get/os.Getenv to try to bypass this function! ⚠️
//   - In older versions this env var was bound to Viper. Viper does not support case-sensitive keys.
//     This means someone could export `IoNoS_API_url` and it would work. So - don't try it. Would be breaking.
//   - Using viper.Get instead of this func means you rely on non-deterministic global behaviour. Which is hard to debug.
//     You won't have the certainty that ArgServerUrl is bound to the chain (ArgServerUrl, EnvServerUrl, CfgServerUrl) when you call viper.Get(ArgServerUrl)
//     so, a very natural reaction to this is checking the next chain variables yourself or leaving it blank - which creates a ton of technical debt and code duplication, or unpredictable behaviour for the user
```

cristiGuranIonos (dismissed):

commands/dns/zone/delete.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+
+			zoneId, err := Resolve(viper.GetString(core.GetFlagName(c.NS, constants.FlagZone)))
+			if err != nil {
+				return err
+			}
+
+			z, _, err := client.Must().DnsClient.ZonesApi.ZonesFindById(context.Background(), zoneId).Execute()
+			if err != nil {
+				return fmt.Errorf("failed getting zone by id %s", zoneId)
```
cristiGuranIonos, line 42: we should propagate error here also?

avirtopeanu-ionos: good catch, thank you!

cristiGuranIonos (dismissed):

cristiGuranIonos (dismissed):

mimihalescu (approved):

PR #301, opened 2023-06-14
feat: ionosctl whoami

avirtopeanu-ionos:
1. Make it easy to find the source (env, config, flag, etc) of each of the credentials
2. Make it possible to find out who we're currently logged in as.
3. If using token, decode it to find user UUID and use the UM API to get the user.

### DIFF
--- a/internal/jwt/token.go
+++ b/internal/jwt/token.go
@@ -41,12 +41,12 @@ func Claims(token string) (map[string]interface{}, error) {
 	payloadDecoded := base64.NewDecoder(base64.StdEncoding, payload)
 	decBytes, err := io.ReadAll(payloadDecoded)
 	if err != nil {
-		return nil, fmt.Errorf("failed to decode payload: %v", err)
+		return nil, fmt.Errorf("failed to decode payload: %w", err)
 	}
 
 	var claims map[string]interface{}
-	if err := json.Unmarshal(decBytes, &claims); err != nil {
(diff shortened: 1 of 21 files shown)


internal/jwt/token.go
```diff
@@ -0,0 +1,89 @@
...
+
+func Claims(token string) (map[string]interface{}, error) {
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("incorrect format of JWT token")
+	}
+
+	payload, err := base64.RawURLEncoding.DecodeString(pad(parts[1]))
+	if err != nil {
+		return nil, fmt.Errorf("could not decode JWT payload: %v", err)
```
cristiGuranIonos, line 30: %w

internal/jwt/token.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	var claims map[string]interface{}
+	if err := json.Unmarshal(payload, &claims); err != nil {
+		return nil, fmt.Errorf("could not parse JWT payload: %v", err)
```
cristiGuranIonos, line 35: %w

avirtopeanu-ionos:
this branch is hard to maintain separately of enhancement/config-creds. I will merge it. We can extract smaller branches from that one afterwards, but for now I'll keep them together to avoid merge conflicts

PR #305, opened 2023-06-20
doc: front matter quotes

avirtopeanu-ionos:
Having multiple "keys" in the description of a command would break the documentation. For example,

```
Error in user YAML: (<unknown>): mapping values are not allowed in this context at line 1 column 35
---
description: Create a record. Wiki: https://docs.ionos.com/dns-as-a-service/readme/api-how-tos/create-a-new-dns-record
---
```

This section is called a "Front Matter", and must provide metadata via key value pairs formatted in valid YAML. 

This PR encloses value of "description" in quotes to stop it from breaking if more things that look like key-value are sent.

```
---
description: "Create a record. Wiki: https://docs.ionos.com/dns-as-a-service/readme/api-how-tos/create-a-new-dns-record"
---
```

### DIFF
--- a/docs/subcommands/Application-Load-Balancer/create.md
+++ b/docs/subcommands/Application-Load-Balancer/create.md
@@ -1,5 +1,5 @@
 ---
-description: Create an Application Load Balancer
+description: "Create an Application Load Balancer"
 ---
 
 # ApplicationloadbalancerCreate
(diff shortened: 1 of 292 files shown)


mimihalescu (approved):

avirtopeanu-ionos:
that last commit was accidentally pushed to this branch, so I removed it

PR #306, opened 2023-06-22
feat: Improve `image upload` user experience

avirtopeanu-ionos:
This PR is based on (uses some functionalities of) #295 

## Overview 

  - Added `--ftp-user`, `--ftp-pass` flags. They can be used to override auth credentials only for the FTP server. Client credentials (API credentials) are now checked only when necessary (e.g. post-upload operations), and are no longer needed to be valid for the initial FTP upload if using these flags.
  - Added a significantly improved documentation and more verbose, detailed errors for this command

### Image already exists: Still try to upload it, but use this info to build a more detailed error 
```
 % ionosctl image upload --location fra --image testtest.vdi
Error: failed while uploading testtest.vdi to FTP server: upload data: cmd failed read expected code 1 with message "": failed to read code, got code 550 and message testtest.vdi: Operation not permitted: 550 testtest.vdi: Operation not permitted
Note: testtest.vdi already exists at ftp-fra.ionos.com. Please contact support at support@cloud.ionos.com to delete the old image. We're sorry for the inconvenience
```
### Using no API credentials (not logged in, in any way) and not passing `--skip-update` 
```
 % ionosctl image upload --location fra --ftp-user $myusername --ftp-pass $mypassword --image testest.vdi 

you did not provide valid API credentials and did not use --skip-update. FTP Upload successful, but cannot query 'GET /images' for your uploaded image: failed selecting an auth layer: none of the layers provided a value for either token or username & password. use `ionosctl whoami --provenance` for help
```
Now you can remain logged in with a JWT, and use these flags for this command without modifying your overall configuration. Also treats the case when you would want to use a custom `--ftp-url`, in this case you definitely do not want to use your IONOS credentials.

### FtpUpload is no longer tied to Image Service

Old: `func (s *imagesService) Upload(ctx context.Context, p UploadProperties) error` 

New: `func FtpUpload(ctx context.Context, p UploadProperties) error`

See above point. If it was to remain tightly coupled to ImagesService, I am forced to initialize a valid ApiClient for it. This is in direct conflict with the command's new help text: `To override your client's (API) credentials, you can use '--ftp-user' and '--ftp-pass', note that if only using only these and not standard auth methods ('ionosctl login'), you may only use this command for FTP uploads.`  (I.e. if having a bad API configuration, the command is still supposed to work in 'upload only mode' with these new flags, but tying this command to Image Service would instead force us to have a correct API configuration).

Note that now the services quite literally no longer serve any purpose and we can safely remove ALL of them, by replacing them all with `client.(apiClient).etc`. 

### New help text

```
OVERVIEW:
  Use this command to securely upload one or more HDD or ISO images to the specified FTP server using FTP over TLS (FTPS). This command supports a variety of options to provide flexibility during the upload process:
  - To override your client's (API) credentials, you can use '--ftp-user' and '--ftp-pass', note that if only using only these and not standard auth methods ('ionosctl login'), you may only use this command for FTP uploads.
  - The command supports renaming the uploaded images with the '--image-alias' flag. If uploading multiple images, you must provide an alias for each image.
  - Specify the context deadline for the FTP connection using the '--timeout' flag. The operation as a whole will terminate after the specified number of seconds, i.e. if the FTP upload had finished but your PATCH operation did not, only the PATCH operation will be intrerrupted.

POST-UPLOAD OPERATIONS:
  By default, this command will query 'GET /images' endpoint for your uploaded images, then try to use 'PATCH /images/<UUID>' to update the uploaded images with the given property fields.
  - It is necessary to use valid API credentials for this.
  - To skip this API behaviour, you can use '--skip-update'.

CUSTOM URLs:
  This command supports usage of other FTP servers too, not just the IONOS ones.
  - The '--location' flag is only required if your '--ftp-url' contains a placeholder variable (i.e. %s).
  In this case, for every location in that slice, an attempt of FTP upload would be made at the URL computed by embedding it into the placeholder variable
  - Use the '--skip-verify' flag to skip the verification of the server certificate. This can be useful when using a custom ftp-url,
  but be warned that this could expose you to a man-in-the-middle attack.
  - If you're using a self-signed FTP server, you can provide the path to the server certificate file using the '--crt-path' flag.

EXAMPLES:
- 'ionosctl img u -i kolibri.iso -l fkb,fra,vit --skip-update': Simply upload the image 'kolibri.iso' from the current directory to IONOS FTP servers 'ftp://ftp-fkb.ionos.com/iso-images', 'ftp://ftp-fra.ionos.com/iso-images', 'ftp://ftp-vit.ionos.com/iso-images'.
- 'ionosctl img u -i kolibri.iso -l fra': Upload the image 'kolibri.iso' from the current directory to IONOS FTP server 'ftp://ftp-fra.ionos.com/iso-images'. Once the upload has finished, start querying 'GET /images' with a filter for 'kolibri', to get the UUID of the image as seen by the Images API. When UUID is found, perform a 'PATCH /images/<UUID>' to set the default flag values.

- 'ionosctl img u -i kolibri.iso --skip-update --skip-verify --ftp-url ftp://12.34.56.78': Use your own custom server. Use skip verify to skip checking server's identity
- 'ionosctl img u -i kolibri.iso -l fra --ftp-url ftp://myComplexFTPServer/locations/%s --crt-path certificates/my-servers-cert.crt --location Paris,Berlin,LA,ZZZ --skip-update': Upload the image to multiple FTP servers, with location embedding into URL.


```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,13 @@
 ## [6.7.0] (XXX 2023)
 
 ### Added
-- Added `cfg logout` command, which iterates through fields in your config file which contain sensitive data and removes their value
-- Added `cfg location` command which prints config file path.
-- Added `cfg whoami` command which prints the authenticated user, with the option (`-p/--provenance`) to print the source and rules/order  of credentials.
-- Added `--ftp-user`, `--ftp-pass` to image upload command, which doesn't support JWT token. They can be used to override auth credentials only for the FTP server.
+- Config:
+  - Added `cfg logout` command, which iterates through fields in your config file which contain sensitive data and removes their value
+  - Added `cfg location` command which prints config file path.
+  - Added `cfg whoami` command which prints the authenticated user, with the option (`-p/--provenance`) to print the source and rules/order  of credentials.
+- Command `image upload`:
+  - Added `--ftp-user`, `--ftp-pass` flags. They can be used to override auth credentials only for the FTP server. Your API credentials are now checked only when necessary (e.g. post-upload operations), and are no longer needed to be valid for the initial FTP upload.
+  - Added a significantly improved documentation and more verbose, detailed errors for this command
 
 ### Changed
 - `ionosctl login` is now deprecated: it will simply call `ionosctl cfg login`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Added `cfg logout` command, which iterates through fields in your config file which contain sensitive data and removes their value
 - Added `cfg location` command which prints config file path.
 - Added `cfg whoami` command which prints the authenticated user, with the option (`-p/--provenance`) to print the source and rules/order  of credentials.
-
+- Added `--ftp-user`, `--ftp-pass` to image upload command, which doesn't support JWT token. They can be used to override auth credentials only for the FTP server.
 
 ### Changed
 - `ionosctl login` is now deprecated: it will simply call `ionosctl cfg login`.
--- a/commands/cloudapi-v6/image.go
+++ b/commands/cloudapi-v6/image.go
@@ -205,6 +205,11 @@ func ImageCmd() *core.Command {
 		CmdRun:     RunImageUpload,
 		InitClient: true,
 	})
+
+	// Username, Password override flags. DO NOT BIND TO VIPER, as the environment variables are linked to some linux things on certain apps, incl. zsh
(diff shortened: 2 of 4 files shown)


cristiGuranIonos (approved):

avirtopeanu-ionos (commented):
note: `ionosctl image upload` doesn't verify IONOS_PINNED_CERT environment variable, users must workaround this by putting it into a file and using the --crt-path. This should be addressed before merging

avirtopeanu-ionos:
#358

PR #322, opened 2023-07-20
fix: `image upload`: hide --image-alias, add --rename

avirtopeanu-ionos:
`--image-alias` if set, renames the uploaded images before upload, so it is confusing for the user that no image alias is set

This PR hides the `--image-alias` flag in the help text, and introduces `--rename` flag. Both of these flags are handled the same in the actual implementation of the command, so there is no change for existing automation that uses ionosctl


avirtopeanu-ionos:
#411

PR #324, opened 2023-07-20
feat: auto generated and random postfix for server names

jonas27:

Create a random postfix of 8 characters for server names. This is implemented with the "--generate-name" flag. The name and implementation are similar to kubernetes.

## Rational

I noticed that names in a datacenter are not unique to resources, which is different from what I come to expect from other platforms (e.g. Kubernetes, Azure, GCP).
It also means that names carry less additional value, e.g. I can't do `ionosctl server get -n test-server-1 --dataceneter-id test-id`.
Using auto-generated postfixes makes this shortcoming more explicit.

To be honest, the optimal solution would be to enforce unique name+resource combinations inside a datacenter on the server. 

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated

### DIFF
--- a/commands/cloudapi-v6/server.go
+++ b/commands/cloudapi-v6/server.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"strconv"
 	"strings"
@@ -31,6 +32,8 @@ import (
 const (
 	serverCubeType       = "CUBE"
 	serverEnterpriseType = "ENTERPRISE"
+
+	generateLenRandomPostfix = 8
 )
 
 func ServerCmd() *core.Command {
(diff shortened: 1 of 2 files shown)


avirtopeanu-ionos:
First of all, getting a resource by name is not yet implemented, specifically because of this name uniqueness issue.

But yes, I fully agree with you, I also want name uniqueness enforced on Cloud API resources, this subject has been discussed so many times, it is so frustrating for me too. I think that your approach is one possible way of handling this on the `ionosctl` side, but the problem that always blocks this is the idea that if these resources are created some place else e.g Terraform, Ansible, DCD, etc, we have to choose a resource for the user. The team, last time we talked about this, was against failing the operation (on `ionosctl`) if more than one resource with the same name is found (like how Ansible does it).  

As for the other APIs which DO enforce resource name uniqueness, we have decided to implement something similar to your "get by name" functionality mentioned above. It will be very similar to how the implementation for the DNS namespace works - eg `--resource` instead of `--resource-id` which can handle both names and UUIDs

jonas27:
> First of all, getting a resource by name is not yet implemented, specifically because of this name uniqueness issue.
> 
> But yes, I fully agree with you, I also want name uniqueness enforced on Cloud API resources, this subject has been discussed so many times, it is so frustrating for me too. I think that your approach is one possible way of handling this on the `ionosctl` side, but the problem that always blocks this is the idea that if these resources are created some place else e.g Terraform, Ansible, DCD, etc, we have to choose a resource for the user. The team, last time we talked about this, was against failing the operation (on `ionosctl`) if more than one resource with the same name is found (like how Ansible does it) - and picking a random resource out of the returned ones is also not great
> 
> As for the other APIs which DO enforce resource name uniqueness, we have decided to implement something similar to your "get by name" functionality mentioned above. It will be very similar to how the DNS commands work - eg `--resource` which can handle both names and UUIDs instead of `--resource-id` which can handle only UUIDs

@avirtopeanu-ionos thanks and agreed. I think both, name postfix generation and uniqueness should be dealt with on the server side as is proper. Hence I am closing this PR again and hope the features are coming soon.

PR #325, opened 2023-07-24
refactor: completions option

avirtopeanu-ionos:
(WIP)

Some flag completions are mistakenly bound to other flags, this PR adds a FlagOption to add completions, so there is a single source of truth for command name and no dev mistakes can be made in the future


avirtopeanu-ionos:
Introduced in #468

Issue #249, opened 2023-03-08
-o json Output is different from API


maboehm:
## Description
The JSON output generated by addding the flag `-o json` to any command is different from the output produced when calling an endpoint directly. E.g.
```
$ ionosctl datacenter list -o json | jq 'keys'
[
  "Resources"
]
$ curl -s https://api.ionos.com/cloudapi/v6/datacenters --user "$IONOS_USERNAME:$IONOS_PASSWORD" | jq keys
[
  "_links",
  "href",
  "id",
  "items",
  "limit",
  "offset",
  "type"
]
```
I don't necessarily mind that the metadata is lost, but the fact that the top level key `items` gets renamed to `Resources` is very inconvenient - e.g. I would have expected that a `jq` query works the same whether I use the API directly or `ionosctl -o json`

## Expected behavior
JSON fields in the API Response are not renamed

## Environment

Ionosctl version: 
```
v6.5.1
```

## References
This is caused by the way resulting lists are passed to the Printer, which specifies the JSON representation to be `Resources` https://github.com/ionos-cloud/ionosctl/blob/master/pkg/printer/result.go#L132

avirtopeanu-ionos:
Will be fixed by #348

avirtopeanu-ionos:
#348 has been merged into master, now all API fields will be correctly shown in `-o json`

```
 % i dc ls -o json | jq keys
[
  "_links",
  "href",
  "id",
  "items",
  "limit",
  "offset",
  "type"
]
```

You can expect this to be included in the next release!

PR #332, opened 2023-07-31
fix: Don't rewrap JSON output

avorima:

The CLI rewraps any output that's not a `Result` in a dummy struct for marshalling.
As a user, I would expect `-o json` to be the "raw" version of the response from the server, i.e. without omitting or adding any extra fields.
Adding some indent is fine, since it's a CLI which is likely used by humans.

Consider this example to get the newest available k8s version:

```
curl ... https://api.ionos.com/cloudapi/v6/k8s/versions | jq -r '. | first'
ionosctl k8s version list -o json | jq -r '.Message | first'
```

It can be somewhat confusing for new users of `ionosctl` if the output differs in such a slight way.

This is somewhat related to https://github.com/ionos-cloud/ionosctl/issues/249, but I have no issues with the `items` approach that was implemented.

## Checklist

- [X] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated

### DIFF
--- a/pkg/printer/printer.go
+++ b/pkg/printer/printer.go
@@ -73,9 +73,7 @@ func (p *JSONPrinter) write(out io.Writer, v interface{}) error {
 			return err
 		}
 	default:
-		var msg DefaultMsgPrint
-		msg.Message = v
-		err := WriteJSON(&msg, out)
+		err := WriteJSON(v, out)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


avirtopeanu-ionos (approved):
Thanks for this PR!

PR #341, opened 2023-08-08
feat: new output package

glimberea:

Adds 2 new packages, one for output generation and the other for JSON to table/text conversion.

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


pkg/json2table/json2table_unit_test.go
```diff
@@ -0,0 +1,128 @@
+//go:build unit
+// +build unit
```
avirtopeanu-ionos, line 2: we don't need annotations for unit tests, only for integration tests.

Also, the file name can be just `json2table_test.go`

However I have no problem with keeping these if you think they are useful

pkg/json2table/json2table_unit_test.go
```diff
@@ -0,0 +1,128 @@
...
+		return false
+	}
+
+	m2, err := json.Marshal(mapSlice2)
+	if err != nil {
+		return false
+	}
+
+	return string(m1) == string(m2)
+}
```
avirtopeanu-ionos, line 128: I'm scared of this func - why not use `assert.Equal` directly, or `reflect.DeepEqual`?

You're marshalling an unordered collection (a map) into a serialized string comparing them. which MIGHT be problematic, I'm not sure if this might introduce any problems or not

avirtopeanu-ionos: or if you really want to compare them in a separate func, maybe compare them key by key ?

pkg/json2table/json2table_unit_test.go
```diff
@@ -0,0 +1,128 @@
...
+			"int32":  testInt32,
+			"string": testStr,
+		},
+	}
+)
+
+func TestConvertJSONToText(t *testing.T) {
+	t.Run("Convert JSON to TEXT with basic struct", testConvertJSONToTextWithBasicStruct)
+	t.Run("Convert JSON to TEXT with complex struct", testConvertJSONToTextWithComplexStruct)
+	t.Run("Convert JSON to TEXT with inner basic structs", testConvertJSONToTextWithInnerBasicStructs)
```
avirtopeanu-ionos, line 95: would be nice to have some negative test cases too, to test if it fails gracefully, e.g:
-  malformed json
- paths to unexpected types (e.g. path to an expected array is actually just a string) and vice versa 
- empty inputs (json / paths)

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	var buff = new(bytes.Buffer)
+	var w = new(tabwriter.Writer)
+	w.Init(buff, 5, 0, 3, ' ', tabwriter.StripEscape)
+
+	if !viper.IsSet(constants.ArgNoHeaders) {
+		if _, err = fmt.Fprintln(w, strings.Join(cols, "\t")); err != nil {
+			return "", nil
```
avirtopeanu-ionos, line 74: This should probably return an error

```suggestion
			return "", fmt.Errorf("failed writing columns: %w", err)
```

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,161 @@
...
+	if outputFormat == "json" {
+		return generateJSONOutput(obj)
+	}
+
+	if outputFormat == "text" {
+		return generateTextOutput("", convertedObj, nil, cols, true)
+	}
+
+	return "", fmt.Errorf(outputFormatErr, outputFormat)
+}
```
avirtopeanu-ionos, line 44: funcs are redundant - they share too much functionality

we should probably create a new private func, which both of these funcs should use

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,161 @@
...
+			return "", err
+		}
+	}
+
+	if err = w.Flush(); err != nil {
+		return "", err
+	}
+
+	return buff.String(), nil
+}
```
avirtopeanu-ionos, line 90: 1. Using booleans as argument is really hard to read e.g what does `foo("bar", true)` do differently vs `foo("bar", false)`? I would consider using a separate function, or a Go Option for generating without conversion
2. This func does too much and should probably be split into multiple helper, private funcs

```suggestion
func generateTextOutputWithConversion(rootPath string, obj interface{}, jsonPaths map[string]string, cols []string) (string, error) {
	table, err := convertToTable(rootPath, jsonPaths, obj)
	if err != nil {
		return "", err
	}
	return writeTableToText(table, cols), nil
}

func generateTextOutputWithoutConversion(convertedObj []map[string]interface{}, cols []string) (string, error) {
	return writeTableToText(convertedObj, cols), nil
}

func convertToTable(rootPath string, jsonPaths map[string]string, obj interface{}) ([]map[string]interface{}, error) {
	return json2table.ConvertJSONToTable(rootPath, jsonPaths, obj)
}

func writeTableToText(table []map[string]interface{}, cols []string) string {
	var buff = new(bytes.Buffer)
	var w = new(tabwriter.Writer)
	w.Init(buff, 5, 0, 3, ' ', tabwriter.StripEscape)

	if !viper.IsSet(constants.ArgNoHeaders) {
		fmt.Fprintln(w, strings.Join(cols, "\t"))
	}

	for _, t := range table {
		format, values := convertTableToText(cols, t)
		fmt.Fprintf(w, format+"\n", values...)
	}

	w.Flush()
	return buff.String()
}
```

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -124,6 +78,48 @@ func GenerateRawOutput(a interface{}) string {
...
+	if err != nil {
+		return "", fmt.Errorf("failed converting object to table using %+v: %w", jsonPaths, err)
+	}
+
+	return writeTableToText(table, cols), nil
+}
+
+func generateTextOutputFromTable(table []map[string]interface{}, cols []string) (string, error) {
+	return writeTableToText(table, cols), nil
+}
```
avirtopeanu-ionos, line 101: this is very nice :100:

pkg/json2table/json2table.go
```diff
@@ -0,0 +1,89 @@
...
+	needsFlatten := true
+	for _, child := range tempChildren {
+		switch child.Data().(type) {
+		case []interface{}:
+			children = append(children, child.Children()...)
+		default:
+			needsFlatten = false
+		}
+
+		if !needsFlatten {
```
cristiGuranIonos, line 82: can't this part be inside the default case?

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+
+func GenerateOutputPreconverted(obj interface{}, convertedObj []map[string]interface{}, cols []string) (string, error) {
+	if viper.IsSet(constants.ArgQuiet) {
+		return "", nil
+	}
+
+	outputFormat := viper.GetString(constants.ArgOutput)
+
+	if outputFormat == "json" {
```
cristiGuranIonos, line 50: JSONFormat?

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func GenerateRawOutput(a interface{}) string {
+	if viper.IsSet(constants.ArgQuiet) {
+		return ""
+	}
+
+	outputFormat := viper.GetString(constants.ArgOutput)
+
+	if outputFormat == "json" {
```
cristiGuranIonos, line 88: JSONFormat

pkg/json2table/json2table.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/Jeffail/gabs/v2"
+)
+
+func ConvertJSONToTable(rootPath string, jsonPaths map[string]string, rootObj interface{}) ([]map[string]interface{}, error) {
+	if rootObj == nil {
+		return nil, fmt.Errorf("object provided cannot be nil")
```
avirtopeanu-ionos, line 12:
```suggestion
		return nil, fmt.Errorf("provided object cannot be nil")
```

pkg/json2table/json2table.go
glimberea: I'm afraid not, because I want to break out of the loop. If I add the break inside the switch, it will be registered as a break between switch cases.

cristiGuranIonos: can't it be changed to an if statement?

pkg/json2table/json2table.go
```diff
@@ -0,0 +1,89 @@
+package json2table
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/Jeffail/gabs/v2"
+)
+
+func ConvertJSONToTable(rootPath string, jsonPaths map[string]string, rootObj interface{}) ([]map[string]interface{}, error) {
```
avirtopeanu-ionos, line 10: Can you add some comments, detailing what each variable should be? Also it would be nice to have more descriptive argument names. For example `rootObj` -> `jsonOrStructObject`/`sourceData`, `rootPath` -> `jsonPathsStartingPoint`/`jsonPathsPrefix`/`rootLocator`

avirtopeanu-ionos:
```suggestion
// ConvertJSONToTable extracts data from a JSON or struct object using specified paths.
//
// jsonPathsPrefix: Points to a specific location in the JSON or struct object from where 
// extraction should begin. If left empty, extraction will start from the root of the object.
//
// columnPathMapping: A map where each key represents the desired column name in the output 
// table, and each value represents a gabs path to extract the data within the given JSON or struct.
//
// sourceData: JSON or struct from which data extraction should take place.
//
// Returns a slice of maps. Each map represents a row in the table, with each key-value 
// pair in the map being equivalent to a column name and its value.
```

pkg/json2table/json2table.go
```diff
@@ -0,0 +1,89 @@
...
+			needsFlatten = false
+		}
+
+		if !needsFlatten {
+			children = tempChildren
+			break
+		}
+	}
+
+	return children, nil
```
avirtopeanu-ionos, line 88:
```suggestion
var children []*gabs.Container

for _, child := range tempChildren {
	switch child.Data().(type) {
	case []interface{}:
		children = append(children, child.Children()...)
	default:
		return tempChildren, nil
	}
}

return children, nil
```

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+		field := table[col]
+		formats = append(formats, "%v")
+
+		if field == nil {
+			values = append(values, "")
+
+			continue
+		}
+
+		switch field.(type) {
```
avirtopeanu-ionos, line 162: (nitpicking) 

You can reduce the repetition of the types here by doing `switch fieldType := field.(type) {`

This would mean you would have less type casts. But the current version is fine too.

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+func generateTextOutputFromJSON(rootPath string, obj interface{}, jsonPaths map[string]string, cols []string) (string, error) {
+	table, err := json2table.ConvertJSONToTable(rootPath, jsonPaths, obj)
+	if err != nil {
+		return "", fmt.Errorf("failed converting object to table using %+v: %w", jsonPaths, err)
+	}
+
+	return writeTableToText(table, cols), nil
+}
+
+func generateTextOutputFromTable(table []map[string]interface{}, cols []string) (string, error) {
```
avirtopeanu-ionos, line 122: it looks like this func never returns an error. Why not use `writeTableToText` directly though?

pkg/json2table/json2table.go
avirtopeanu-ionos: See my suggestion at https://github.com/ionos-cloud/ionosctl/pull/341#discussion_r1332769444 maybe? Won't that work?

glimberea: It can be replaced with an if statement, so I'll be doing that, combined with the direct return from https://github.com/ionos-cloud/ionosctl/pull/341#discussion_r1332769444. Thanks for the idea

pkg/jsontabwriter/jsontabwriter.go
glimberea: tbh, I don't remember, but I will change it :))))

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+func GenerateRawOutput(a interface{}) string {
+	if viper.IsSet(constants.ArgQuiet) {
+		return ""
+	}
+
+	outputFormat := viper.GetString(constants.ArgOutput)
+
+	if outputFormat == "json" {
+		out, err := generateJSONOutput(a)
+		if err != nil {
```
cristiGuranIonos, line 90: if err != nil, generateJSONOutput returns "". So in this case, just return out if we don't care about the error? Maybe comment why we don't care about the error.

pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -0,0 +1,204 @@
...
+		format, values := convertTableToText(updatedCols, t)
+		fmt.Fprintf(w, format+"\n", values...)
+	}
+
+	w.Flush()
+
+	return buff.String()
+}
+
+func convertTableToText(cols []string, table map[string]interface{}) (format string, values []interface{}) {
```
cristiGuranIonos, line 148: should this be formats?

avirtopeanu-ionos (approved):
Looks fantastic!!! :100: :100:

PR #348, opened 2023-08-25
refactor: replace printer pkg

glimberea:
The old `Printer` package needs to be replaced with the new `jsontabwriter` and `json2table` output generation packages, so that `stdout` and `stderr` for a command are no longer hard coded.

TODO:
- [x] replace in DBaaS commands
- [x] replace `confirmation.AskForConfirm (deprecated)` with `confirm.Ask`
- [x] replace `clierror.CheckError (deprecated)` with `die.Die`
- [x] fix/reevaluate dataplatform nodepool printing (storage and ram cols)
- [ ] test for errors


commands/certmanager/apiversion.go
```diff
@@ -26,63 +35,31 @@ func CertGetApiVersionCmd() *core.Command {
 		InitClient: true,
 	})
 
-	cmd.Command.Flags().StringSlice(constants.ArgCols, nil, printer.ColsMessage(allCols))
+	cmd.Command.Flags().StringSlice(constants.ArgCols, nil, tabheaders.ColsMessage(defaultCertificateCols))
```
avirtopeanu-ionos, line 38: `defaultCertificateCols` is printed as the set of all available columns, yet `allApiVersionCols` used for autocompleting the available columns

commands/certmanager/delete.go
```diff
@@ -46,36 +48,42 @@ func CertDeleteCmd() *core.Command {
...
 		if err != nil {
 			return err
 		}
+
 		for _, cert := range *certs.Items {
 			msg := fmt.Sprintf("delete Certificate ID: %s", *cert.Id)
-			if err := utils.AskForConfirm(c.Stdin, c.Printer, msg); err != nil {
-				return err
+			if !confirm.FAsk(c.Command.Command.InOrStdin(), msg, viper.GetBool(constants.ArgForce)) {
+				return nil
```
avirtopeanu-ionos, line 68: Is this intended? i.e. no longer throwing an error if the user refuses the confirmation

commands/cloudapi-v6/applicationloadbalancer_rule_test.go
```diff
@@ -489,6 +491,7 @@ func TestRunApplicationLoadBalancerForwardingRuleDeleteAllListErr(t *testing.T)
 		viper.Set(core.GetFlagName(cfg.NS, cloudapiv6.ArgApplicationLoadBalancerId), testAlbForwardingRuleVar)
 		viper.Set(core.GetFlagName(cfg.NS, cloudapiv6.ArgAll), true)
 		rm.CloudApiV6Mocks.ApplicationLoadBalancer.EXPECT().ListForwardingRules(testAlbForwardingRuleVar, testAlbForwardingRuleVar, gomock.AssignableToTypeOf(testListQueryParam)).Return(testAlbForwardingRules, nil, testApplicationLoadBalancerForwardingRuleErr)
+		cfg.Command.Command.SetIn(bytes.NewReader([]byte("\n")))
```
avirtopeanu-ionos, line 494: Just an observation that this test already uses `--force` and this isn't needed

commands/cloudapi-v6/completer/properties.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // DatacenterCPUFamilies returns the list of available CPU families in a given datacenter
 // If the datacenter ID is not provided returns a set of possible default values.
-func DatacenterCPUFamilies(ctx context.Context, outErr io.Writer, datacenterId string) []string {
+func DatacenterCPUFamilies(ctx context.Context, _ io.Writer, datacenterId string) []string {
```
avirtopeanu-ionos, line 14: Can safely remove the unused argument

commands/dbaas/mongo/cluster/delete.go
```diff
@@ -64,17 +65,17 @@ ionosctl db m c d --all --name <name>`,
...
+				keepGoing := confirm.FAsk(c.Command.Command.InOrStdin(), fmt.Sprintf("%s, try deleting %s anyways", wrapped.Error(), clusterId))
 				if !keepGoing {
 					return wrapped
 				}
 			}
 
-			ok := confirm.Ask(confirmStringForCluster(chosenCluster), viper.GetBool(core.GetFlagName(c.NS, constants.ArgForce)))
+			ok := confirm.FAsk(c.Command.Command.InOrStdin(), confirmStringForCluster(chosenCluster), viper.GetBool(constants.ArgForce))
 			if !ok {
 				return fmt.Errorf("user denied confirmation")
```
avirtopeanu-ionos, line 76: For some resources we throw an error if user denies confirmation, yet for others we don't.

Let's make a decision and update them all to be the same

commands/version.go
```diff
@@ -41,14 +42,11 @@ func VersionCmd() *core.Command {
...
 func RunVersion(c *core.CommandConfig) error {
-	err := c.Printer.Print(rootCmd.Command.Version)
-	if err != nil {
-		return err
-	}
-	c.Printer.Verbose("sdk-go " + sdkcompute.Version)
-	c.Printer.Verbose("sdk-go-dbaas-postgres " + sdkpostgres.Version)
-	c.Printer.Verbose("sdk-go-auth " + sdkauth.Version)
-	c.Printer.Verbose("sdk-go-cert-manager " + sdkcertmanager.Version)
+	fmt.Fprintf(c.Command.Command.OutOrStdout(), jsontabwriter.GenerateLogOutput(c.Command.Command.Version))
```
avirtopeanu-ionos, line 45: This doesn't work for some reason.

```
 % i version

```

commands/certmanager/delete.go
glimberea: yes, it was intentional given that `confirm.FAsk` never throws an error (even when the user inputs gibberish instead of yes/no). i figured it was intentionally thought out to be that way, but it would be easy to add a check for invalid input and have it return an error. the return value for when `no` should still be `nil` though.

commands/version.go
avirtopeanu-ionos: Found it

```suggestion
	fmt.Fprintf(c.Command.Command.OutOrStdout(), jsontabwriter.GenerateLogOutput(rootCmd.Command.Version))
```

But wow, having `Version` defined on every single command is truly... something. We should get rid of that.

glimberea: 🤦🏻‍♀️ 
yeah, we really should get that fixed

commands/certmanager/delete.go
avirtopeanu-ionos: well, initially the CLI would throw errors if the user inputs gibberish / no. `user denied confirmation`

However, I don't mind this change, because:
1. These errors are not really part of any contract / promise
2. These errors are not intended for software integrations, etc, since those users generally want to use `-f/--force` and never have to deal with them
3. I can't think of any use case in which the user would want to parse the resulting output of the API in case he answers `no` upon changing his mind about the deletion / accidentally running the command.

So, I don't necessarily like this change, but I'm okay with it - just have to adjust the other places where an err is returned and return nil instead.

glimberea: Since you're not 100% on board with this change, I will make it have the same behavior as before (error on no/gibberish), just so we are consistent and we don't cause any potential breaking changes for users (I don't know who would parse the error output in this case, but it's better to be safe than sorry). :))))

avirtopeanu-ionos (approved):
Great work! :1st_place_medal:

PR #350, opened 2023-08-29
fix #349: api url env var

### LINKED ISSUE #349: IONOS_API_URL not working since v6.6.4
## Description

When using ionosctl v6.6.4 or newer (tested up until 6.6.8), the `IONOS_API_URL` environment variable does not seem to be used anymore.
Instead of accessing the api endpoint configured this way, the variable is ignored and the default value used.

## Expected behavior

I expect the environment variable `IONOS_API_URL` to define the API endpoint accessed by ionosctl, similar to the `--api-url` / `-u` flag.

## Environment

Ionosctl version:
```
6.6.4 (and newer)
```
OS:
```
darwin/arm64
```
Shell:
```
zsh
```

## How to Reproduce

1. Set up `ionosctl` for normal API access, e.g. by exporting the `IONOS_USERNAME` and `IONOS_PASSWORD` env variables
2. Verify that API access works using `ionosctl dc list`
3. Set `export IONOS_API_URL=doesnt.exist` or a similar url that should not work
4. Run `ionosctl dc list`. Observe that the same result as in step 2 is obtained.
5. (optional) run `ionosctl -u $IONOS_API_URL dc list` and observe an error

avirtopeanu-ionos:
fixes #349: `IONOS_API_URL` being overriden by default flag values

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -49,7 +49,6 @@ func GetServerUrl() string {
 	}
 	if cfgVal := viper.GetString(constants.ServerUrl); viper.IsSet(constants.ServerUrl) {
 		// 3. Fallback to non-empty cfg field
-		fmt.Println("cfg val: " + cfgVal)
 		return cfgVal
 	}
 	// 4. Return empty string. SDKs should handle it, per docs
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -9,9 +9,9 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/ionos-cloud/ionosctl/v6/pkg/constants"
-
 	cloudv6 "github.com/ionos-cloud/sdk-go/v6"
 
 	"github.com/spf13/viper"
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -36,15 +36,18 @@ func GetUserData() map[string]string {
 //	  so, a very natural reaction to this is checking the next chain variables yourself or leaving it blank - which creates a ton of technical debt and code duplication, or unpredictable behaviour for the user
 func GetServerUrl() string {
 	viper.AutomaticEnv()
-	if flagVal := viper.GetString(constants.ArgServerUrl); flagVal != "" {
-		// 1. Above all, use if the global flag is set
-		return flagVal
+	if flagVal := viper.GetString(constants.ArgServerUrl); viper.IsSet(constants.ArgServerUrl) {
+		// 1. Above all, use global flag val
+		if flagVal != "dns.de-fra.ionos.com" {
```
cristiGuranIonos, line 41: maybe you can use !strings.Contains and use dns.DefaultIonosServerUrl instead of defining here "dns.de-fra.ionos.com" .

cristiGuranIonos (approved):

pkg/config/config.go
avirtopeanu-ionos: can't use that constant because it would lead to circular imports :( but yeah, I will move it to our constants pkg so I can use it here.

PR #352, opened 2023-08-31
refactor: client, config pkgs

avirtopeanu-ionos:
(This is part of an effort to split https://github.com/ionos-cloud/ionosctl/pull/295 into multiple PRs)

Refactors config pkg to introduce an intuitive API `Read (-> map[string]string)` and `Write(map[string]string)` to interact with the config file

Refactors the client pkg to have code that actually represents the layered approach of the CLI for authentication


services/dbaas-postgres/resources/client_test.go
avirtopeanu-ionos, line 1: Duplicated code, already tested in `pkg/config_test.go`

services/dbaas-postgres/resources/backup_test.go
avirtopeanu-ionos, line 1: Making a call to API with a foo variable (passing `test-resource` as UUID) and expecting any error will fail exactly NEVER and it makes no sense to test this. These tests will always pass. Hence all `services/*/resources/*_test.go` have been removed

internal/client/client.go
```diff
@@ -9,91 +10,68 @@ import (
...
+func selectAuthLayer(layers []Layer) (values map[string]string, usedLayer Layer, err error) {
+	for _, layer := range layers {
+		token := viper.GetString(layer.TokenKey)
+		username := viper.GetString(layer.UsernameKey)
+		password := viper.GetString(layer.PasswordKey)
+
+		// TODO (Ask the team for feedback):
+		// If token is set and invalid, we could still try the user & pass here, and choose them over the invalid token.
+		// However this could be inconsistent with the Ansible / TF approach and could confuse the user.
+		// TODO END
```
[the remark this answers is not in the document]
avirtopeanu-ionos: great, I was thinking the same

internal/client/client.go
```diff
@@ -9,91 +10,68 @@ import (
...
+		if layer.TokenKey != "" && token != "" ||
```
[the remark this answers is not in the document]
avirtopeanu-ionos: > do we need to check layer.TokenKey != "" ?

I was thinking that a valid use case of this new layered approach would be to define a layer with only `Token` or only `Username & Password support, in which case the unsupported key for the layer would be empty. 

That being said, yeah, these checks are redundant because `viper.GetString("")` would still return an empty string - so I see your point and we can totally remove these checks.
 
> Can't you just check err above for GetString and only set token if err!= nil? This way here you can just check token?

I'm not entirely sure what you mean here, `GetString` doesn't return an error, but yeah, we can skip the layer.<Token/User>Key checks.

commands/login.go
```diff
@@ -130,7 +130,7 @@ func RunLoginUser(c *core.CommandConfig) error {
 
 	// Store credentials
 	c.Printer.Verbose("Storing credentials to the configuration file: %v", viper.GetString(constants.ArgConfig))
-	err = config.WriteFile()
+	// err = config.WriteFile() // TODO: Replace me in a separate PR
```
mimihalescu, line 133: Is this already covered in another PR?

mimihalescu (approved):

commands/login.go
avirtopeanu-ionos: yeah, in #353 

the line has been commented because `WriteFile` has been replaced by `Write(data)`, and I was too lazy to fix it because:
a) the target branch is not `master` - it is a holding branch `feat/config-rework`, which is not meant to be merged until the whole feature is working correctly
b) this whole login functionality was meant to be reworked anyway (by #353) - so any refactor / fixing on this current version is a waste of time

avirtopeanu-ionos: but i can for sure fix this function call if we want full transparency here

mimihalescu: > but i can for sure fix this function call if we want full transparency here

no need, looks good 👍🏻  thank you for clarifying!

cristiGuranIonos (approved):

PR #354, opened 2023-09-05
feat: whoami

avirtopeanu-ionos:
Introduces a command for debugging authentication:

Valid credentials - Username is printed. In case of Token auth, the JWT is parsed and Users API used to find the username.
```
 % i cfg whoami   
alex[...]@[...].com
```

No credentials provided:
```
 % i whoami    
Note: Authentication failed! None of the authentication layers had a token, or both username & password set.
Authentication layers, in order of priority:
  [1] Global Flags (--token)
  [2] Environment Variables (IONOS_TOKEN, IONOS_USERNAME, IONOS_PASSWORD)
  [3] Config file settings (userdata.token, userdata.name, userdata.password)
```

Invalid credentials - shows used layer and whether using token or username & password:
```
 % i whoami
Note: Authentication failed!
Authentication layers, in order of priority:
  [1] Global Flags (--token)
* [2] Environment Variables (IONOS_TOKEN, IONOS_USERNAME, IONOS_PASSWORD) (USED)
    - Using token for authentication.
    - Using https://api.ionos.com as the API URL.
  [3] Config file settings (userdata.token, userdata.name, userdata.password)
```

Valid credentials and `--provenance`:
```
 % i cfg whoami -p  
Authentication layers, in order of priority:
  [1] Global Flags (--token)
* [2] Environment Variables (IONOS_TOKEN, IONOS_USERNAME, IONOS_PASSWORD) (USED)
    - Using username and password for authentication.
    - Using https://api.ionos.com as the API URL.
  [3] Config file settings (userdata.token, userdata.name, userdata.password)
```

### DIFF
--- a/commands/cfg/whoami.go
+++ b/commands/cfg/whoami.go
@@ -41,13 +41,15 @@ ionosctl cfg whoami --provenance`,
 			if fn := core.GetFlagName(c.NS, FlagProvenance); authErr != nil || viper.GetBool(fn) {
 				var out string
 				if authErr != nil {
-					out = "Note: Authentication failed!\n"
+					out = "Note: Authentication failed!"
+					if cl.UsedLayer() == nil {
+						out += " None of the authentication layers had a token, or both username & password set."
+					}
+					out += "\n"
 				}
 				out += "Authentication layers, in order of priority:\n"
 				for i, layer := range client.ConfigurationPriorityRules {
-					fmt.Println(cl.UsedLayer)
-					fmt.Println(layer)
-					if layer == cl.UsedLayer {
+					if cl.UsedLayer() != nil && *cl.UsedLayer() == layer {
 						out += fmt.Sprintf("* [%d] %s (USED)\n", i+1, layer.Description)
 						if cl.IsTokenAuth() {
 							out += "    - Using token for authentication.\n"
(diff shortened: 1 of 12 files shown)


commands/cfg/whoami.go
```diff
@@ -0,0 +1,90 @@
...
+			if !cl.IsTokenAuth() {
+				// Handle Username & Password Authentication
+				_, err = fmt.Fprintln(c.Command.Command.OutOrStdout(), cl.CloudClient.GetConfig().Username)
+				return err
+			}
+
+			// Handle token authentication
+			usernameViaToken, jwtParseErr := jwt.Username(cl.CloudClient.GetConfig().Token)
+			if jwtParseErr != nil {
+				return fmt.Errorf("failed getting username via token: %w", err)
```
cristiGuranIonos, line 50: err, or jwtParseErr?

cristiGuranIonos (approved):

commands/cfg/whoami.go
avirtopeanu-ionos: yes `jwtParseErr`.

`err` is handled above - it forces `--provenance` if getting credentials fails.

However, I think I will rename `err` to `authErr` for clarity.

avirtopeanu-ionos: ah, sorry, now I see that you meant the question for the embedded error message in `Errorf` call - and not the `if` statement.

Yes, here should be `jwtParseErr` too - good catch

PR #364, opened 2023-09-19
fix: MongoDB SSD bad request error fix

avirtopeanu-ionos:
Fixes a bug with DBaaS Mongo subcommands where creating a default playground mongo cluster would result in a 400 Bad Request `"SSD" is not one of "SSD Standard", "SSD Premium", "HDD"`


commands/dbaas/mongo/cluster/create.go
```diff
@@ -267,7 +267,7 @@ func ClusterCreateCmd() *core.Command {
 	cmd.AddStringFlag(constants.FlagStorageType, "", "SSD",
 		"Custom Storage Type. (only settable for enterprise edition)")
 	_ = cmd.Command.RegisterFlagCompletionFunc(constants.FlagStorageType, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"HDD", "SSD", "\"SSD Premium\""}, cobra.ShellCompDirectiveNoFileComp
+		return []string{"HDD", "\"SSD Standard\"", "\"SSD Premium\""}, cobra.ShellCompDirectiveNoFileComp
```
cristiGuranIonos, line 270: annoying that we don't have a String() function to be able to use mongo.STORAGETYPE_SSD_STANDARD directly

avirtopeanu-ionos: well, I could convert it using go's `string(STORAGETYPE_EXPLICIT_TYPE_HERE)` - but these completions need escaped quotations `\"` because of their spaces

glimberea (approved):

PR #368, opened 2023-09-26
improvement: added new output format

glimberea:
Added `api-json` as a possible output format, which will generate API-like json output. The `json` output format is the same as it was before.


pkg/jsontabwriter/jsontabwriter.go
```diff
@@ -9,16 +9,20 @@ import (
...
 	outputFormatErr = "invalid format: %s"
 
-	// JSONFormat defines the JSON format.
+	// JSONFormat defines the legacy JSON format. It will eventually be removed and fully replaced by APIFormat
+	// (in terms of behavior, the name will be kept)
 	JSONFormat = "json"
 	// TextFormat defines a human-readable format.
 	TextFormat = "text"
+	// APIFormat defines the API matching JSON format. This will be removed once its behavior will be moved to JSONFormat
+	APIFormat = "api-json"
```
avirtopeanu-ionos, line 25: The `-o/--out` possible formats also need to be updated:

```
 % i server list --all -o api-json
Error: unknown type format api-json. Hint: use --output json|text
```

avirtopeanu-ionos (approved):

PR #385, opened 2023-10-17
feat: dynamic completer descriptions

glimberea:
New package for adding descriptive completions more easily, using the `json2table` pkg

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


pkg/completions/completions_test.go
```diff
@@ -0,0 +1,108 @@
...
+		"string":  "string",
+		"float64": "float64",
+	}
+
+	testStr     = "test-string"
+	testInt32   = int32(123)
+	testFloat64 = float64(12.5)
+
+	testInnerStruct = innerStruct{
+		Str:     &testStr,
```
avirtopeanu-ionos, line 34: just a note that you can use `pointer.From("test-string")` to avoid declaring these variables before usage

pkg/completions/completions_test.go
```diff
@@ -0,0 +1,108 @@
...
+func testCompleterAddInfoFail(t *testing.T) {
+	completer := completions.NewCompleter(testConvertedStruct, "int32").AddInfo("stringz")
+	assert.Equal(t, completer, completions.Completer{})
+}
+
+func testCompleterAddInfoWithFormat(t *testing.T) {
+	completer := completions.NewCompleter(testConvertedStruct, "int32").AddInfo("string", "(%v)")
+	assert.NotEqual(t, completer, completions.Completer{})
+
+	printCompleter(completer)
```
avirtopeanu-ionos, line 89: Can you also test the expected result, I for one frankly don't quite understand the intended output here when given the `additionalFormatting` parameter

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [v6.7.1] (???)
 
+## Added
+
+* Added new `completions` package for dynamically generating descriptive completions
```
avirtopeanu-ionos, line 7: So far, we've only used the changelog for changes that are visible to the user. I'm not sure in this case if we should add this, however I have no problem with keeping it in though

avirtopeanu-ionos (approved):

PR #403, opened 2023-11-16
feat: logging service

glimberea:
## Checklist

- [ ] PR name added as appropriate (e.g. `feat:`/`fix:`/`doc:`/`test:`/`refactor:`)
- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


commands/logging-service/logging-service.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/ionos-cloud/ionosctl/v6/internal/core"
+	"github.com/spf13/cobra"
+)
+
+func LoggingServiceCmd() *core.Command {
+	cmd := &core.Command{
+		Command: &cobra.Command{
+			Use:              "logging-service",
+			Aliases:          []string{"log-svc"},
+			Short:            "LaaS Operations. Manage and centralize your application/infrastructure's logs",
```
mimihalescu, line 15:
```suggestion
			Short:            "Logging Service Operations. Manage and centralize your application/infrastructure's logs",
```

commands/logging-service/logs/add.go
```diff
@@ -0,0 +1,124 @@
...
+		constants.FlagLoggingPipelineId,
+		func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			return completer.LoggingServicePipelineIds(), cobra.ShellCompDirectiveNoFileComp
+		},
+	)
+	cmd.AddStringFlag(
+		constants.FlagLoggingPipelineLogTag, "", "", "Sets the tag for the pipeline log", core.RequiredFlagOption(),
+	)
+	cmd.AddSetFlag(
+		constants.FlagLoggingPipelineLogSource, "", "", []string{"docker", "systemd", "generic", "kubernetes"},
```
cbeti-ionos, line 46: this is minor, but since these values are used in multiple places across the logging-service package, we could create some constants for them.

commands/logging-service/logs/add.go
```diff
@@ -0,0 +1,124 @@
...
+		constants.FlagLoggingPipelineLogProtocol, "", "", []string{"http", "tcp"},
+		"Sets the protocol for the pipeline log", core.RequiredFlagOption(),
+	)
+	cmd.AddStringSliceFlag(constants.FlagLoggingPipelineLogLabels, "", nil, "Sets the labels for the pipeline log")
+	cmd.AddStringFlag(
+		constants.FlagLoggingPipelineLogType, "", "loki",
+		"Sets the destination type for the pipeline log",
+	)
+	cmd.AddSetFlag(
+		constants.FlagLoggingPipelineLogRetentionTime, "", "30", []string{"7", "14", "30"},
```
cbeti-ionos, line 59: similar as above

commands/logging-service/logs/get.go
```diff
@@ -0,0 +1,109 @@
...
+
+	for _, l := range *pipeline.Properties.Logs {
+		if *l.Tag == tag {
+			log = l
+
+			break
+		}
+	}
+
+	logConverted, err := resource2table.ConvertLoggingServicePipelineLogToTale(log)
```
cbeti-ionos, line 83: PipelineLogToTable

commands/logging-service/pipeline/create.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+	cmd.AddSetFlag(
+		constants.FlagLoggingPipelineLogRetentionTime, "", "30", []string{"7", "14", "30"},
+		"Sets the retention time in days for the pipeline log",
+	)
+
+	return cmd
+}
+
+func runCreateCmd(c *core.CommandConfig) error {
```
cbeti-ionos, line 64: If I understand correctly, unless the the name of the pipeline is specified, the default behavior is to create an empty pipeline ?

glimberea: No, there are 3 ways to use the create command:

- with the `--json-properties-example` flag, which will print an example for the json properties file that should be used for creating a pipeline (it doesn't actually run the create command)
- with the `--json-properties` flag, which takes the path to a file where the properties of the pipeline have already been defined (the parsing of the file is done in a separate function that was not added by this PR)
- with the required flags (`--name`, `--log-tag`, `--log-source`, `--log-protocol`) and there are some flags which can be populated, but are not required. This will create a pipeline with **only** 1 log

I did not, however, take into consideration the possibility of a user setting both `--json-properties` and `--name`, for instance, so I will look into making that check more clear

PR #404, opened 2023-11-22
feat: implement interactive shell

avirtopeanu-ionos:
Adds an interactive shell via [our new comptplus library](https://github.com/ionoscloudsdk/comptplus/).

image

image

Necessary changes:
- Added a strategy pattern for our confirm package.
    - if used within the `interactive shell`, confirmation messages will be replaced with ` (message)? [to confirm, please use --force]` and then fail the confirmation check rather than ask for user input
    - This is because reading from keyboard within the interactive shell is currently unsupported

- Modified `--filters` behaviour to continue working after being reset without breaking changes for existing usages
    - Only noticeable change is that e.g. `image list --filters ,` would no longer throw an error - but rather work as if `--filters` isn't set at all.


commands/shell.go
```diff
@@ -0,0 +1,89 @@
...
+	ShowHiddenCommands:       true,
+	ShowHiddenFlags:          true,
+	GoPromptOptions: []prompt.Option{
+		prompt.OptionTitle("ionosctl"),
+		prompt.OptionPrefix("> "),
+		prompt.OptionShowCompletionAtStart(),
+
+		prompt.OptionAddKeyBind(prompt.KeyBind{Key: prompt.ShiftLeft, Fn: prompt.GoLeftWord}),
+		prompt.OptionAddKeyBind(prompt.KeyBind{Key: prompt.ShiftRight, Fn: prompt.GoRightWord}),
+		prompt.OptionAddKeyBind(prompt.KeyBind{Key: prompt.ShiftDown, Fn: prompt.GoLineBeginning}),
```
glimberea, line 28: I think the shortcuts are a bit counter-intuitive. You would expect `CTRL+Left/Right` to allow you to move from one word to another, and for `Home` and `End` to get you to the beginning/end of the line (these are functional within the shell at the moment, but still).

commands/shell.go
```diff
@@ -0,0 +1,89 @@
...
+		prompt.OptionAddKeyBind(prompt.KeyBind{Key: prompt.ShiftUp, Fn: prompt.GoLineEnd}),
+		prompt.OptionAddKeyBind(prompt.KeyBind{Key: prompt.ShiftDelete, Fn: prompt.DeleteWord}),
```
glimberea, line 30: Same thing as above, but I think this might require a bit more work to "fix". You'd expect `Delete` to erase characters to the right of the cursor, not the left. From what I saw, there is no `ShiftBackspace` or `ControlBackspace` so that the shortcut could fit the functionality. Perhaps we could implement a replacement for `prompt.DeleteWord` so that the word to the right of the cursor will be deleted?

commands/shell.go
avirtopeanu-ionos: sadly `Home` and `End` don't work within the shell instance, I agree with `CTRL + Left / Right` though.

glimberea: Why don't `Home` and `End` work? I replaced them locally and it worked alright.

avirtopeanu-ionos: on zsh `?[23~` and `?[24~` are entered instead of their respective funcs being executed. However we can just leave the default go-prompt controls I guess, and return to this custom controls topic at a later date when we can allow the user to customize his controls.

commands/shell.go
avirtopeanu-ionos: We can just leave the default go-prompt controls I guess, and return to this custom controls topic at a later date when we can allow the user to customize his controls.

glimberea:
needs https://github.com/ionos-cloud/ionosctl/pull/405 in order to be fully functional

commands/shell.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		err := sliceValue.Replace([]string{})
+		if err != nil {
+			flag.Value.Set(flag.DefValue)
+		}
+	},
+}
+
+type pleaseUseForceInsteadConfirmer struct {
```
cristiGuranIonos, line 73: very polite struct name

glimberea (approved):
Looks good to me 👍🏻

avirtopeanu-ionos:
Fixed a bug where CTRL+C would no longer work after spawning a go-prompt session, however now https://github.com/elk-language/go-prompt is used.

@glimberea can you please take a look again with the new changes?

mimihalescu (approved):

PR #409, opened 2023-12-04
fix #406: empty result should return a zero status code

### LINKED ISSUE #406: bug: `ionosctl` errors with exit code 1 if no images where found
## Description

If you search/filter for a name that does not exists - the tool errors:
```bash
ionosctl image list --filters name=debian-12-amd64-latest-none-exist,location=txl,location=fra
echo $? # Outputs 1
```
This is and leads to unexpected behavior. A command should only error if there is anything wrong with the command/network/filesystem - not if the command return its expected outcome. For example in the CI. If there is no image the tool should output nothing and exit with 0. Or at leas there should be a flag for that.

Please apply to standard behavior

## Expected behavior

Command returns nothing (or empty array if `json` output) and exits with code 0

## Environment

Go Version:
```
?
```
Ionosctl version:
```
ionosctl version v6.7.1
```
OS:
```
Windows 11 Pro 23H2
```
Shell:
```
Git Bash
```

## Configuration Files

None

## How to Reproduce

If you search/filer for a name that does not exists - the tool errors:
```bash
ionosctl image list --filters name=debian-12-amd64-latest-none-exist,location=txl,location=fra
echo $? # Outputs 1
```
This is and leads to unexpected behavior. A command should only error if there is anything wrong with the command/network/or - not if the command return its expected outcome. For example in the CI. If there is no image the tool should output nothing and exit with 0. Or at leas thee should be a flag for that.

Please apply to standard behavior

## Error and Debug Output

```bash
ionosctl image list --filters name=debian-12-amd64-latest,location=txl,location=fra
Error: error getting images based on given criteria
```

## Additional Notes

It is Industry standard to use non 0 exit codes if there is anything wrong - hears it behaves as expected. Most other tools also just exit with 0 when used with filter. See `docker ps --filter name=none-exist` or `kubectl`

Also if you specify `--output json` it should also output an empty array/object. In case of an actual error - the output should also be `json` not text

avirtopeanu-ionos:
Fixes #406: Getting no results for images, requests should return a zero status code.

### DIFF
--- a/commands/cloudapi-v6/image.go
+++ b/commands/cloudapi-v6/image.go
@@ -761,12 +761,6 @@ func RunImageList(c *core.CommandConfig) error {
 		images = sortImagesByTime(images, viper.GetInt(core.GetFlagName(c.NS, cloudapiv6.ArgLatest)))
 	}
 
-	if itemsOk, ok := images.GetItemsOk(); ok && itemsOk != nil {
-		if len(*itemsOk) == 0 {
-			return errors.New("error getting images based on given criteria")
-		}
-	}
-
 	cols := viper.GetStringSlice(core.GetFlagName(c.Resource, constants.ArgCols))
 
 	out, err := jsontabwriter.GenerateOutput("items", jsonpaths.Image, images.Images,
--- a/commands/cloudapi-v6/image.go
+++ b/commands/cloudapi-v6/image.go
@@ -761,6 +761,10 @@ func RunImageList(c *core.CommandConfig) error {
 		images = sortImagesByTime(images, viper.GetInt(core.GetFlagName(c.NS, cloudapiv6.ArgLatest)))
 	}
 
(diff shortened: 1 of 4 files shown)


commands/cloudapi-v6/image.go
```diff
@@ -761,12 +761,6 @@ func RunImageList(c *core.CommandConfig) error {
 		images = sortImagesByTime(images, viper.GetInt(core.GetFlagName(c.NS, cloudapiv6.ArgLatest)))
 	}
 
-	if itemsOk, ok := images.GetItemsOk(); ok && itemsOk != nil {
```
glimberea, line 764: I think a check for `ok == false` and`itemsOk == nil` could still be useful though. Maybe just eliminate the `len` check?

glimberea (approved):
👍🏻

PR #419, opened 2024-01-29
test: add bats e2e tests

avirtopeanu-ionos:
Adds bats-core end to end tests.

- Add libs for simpler asserts
- Adds a complex test for compute resources of CloudAPI and a test for K8s Nodepools
- Adds tagging system, for running only relevant tests
    - Adjust the CI for this tagging system, manually checkout master
    - If failed getting BASE_BRANCH (i.e. like running locally) - then simply choose `master` as the base 
- Retry command for flaky deletions. Ideally our -w flag would be a wait-for-delete on deletion commands, but it's a good compromise for now.


test/bats/setup.bats
```diff
@@ -0,0 +1,58 @@
...
+                sleep $delay;
+            else
+                fail "$n attempts exhausted for command: $*"
+                break
+            fi
+        }
+    done
+}
+
+find_or_create_resource() {
```
glimberea, line 41: Is there a way in which we could make this function accept resources that have plain numeric ids as well? I think it would be useful for testing scenarios which require, for example, preexisting LANs.

avirtopeanu-ionos: Yeah we could probably just get rid of that `assert_regex` statement, and the callee should check whatever it wants to check

glimberea (approved):
Looks good to me, very nice!

PR #422, opened 2024-02-29
fix: Allow disabling node pool autoscaling

avorima:

Fix disabling autoscaling on a Kubernetes node pool. Autoscaling is disabled when the min and max counts are `null`.

Previously this resulted in this error:

```
Error: 422 Unprocessable Entity {
  "httpStatus" : 422,
  "messages" : [ {
    "errorCode" : "110",
    "message" : "[(root).properties.autoScaling.minNodeCount] Attribute has to be greater than 0"
  } ]
}
```

## Checklist

- [ ] Tests added or updated
- [x] Documentation updated
- [ ] Sonar Cloud Scan
- [x] Changelog updated and version incremented (label: upcoming release)
- [ ] Jira task updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Upcoming
+
+### Fixed
+- Kubernetes node pool auto scaling can be disabled by setting `--min-node-count` and `--max-node-count` to 0.
+
 ## [v6.7.4] (January 2024)
 
 ### Added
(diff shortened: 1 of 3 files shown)


avirtopeanu-ionos (approved):
Awesome PR, thank you very much, I like this approach. Also thanks for generating the documentation and adding to the changelog, very thorough work!

Regarding the sonar code quality gate, we will look into that separately, I'm pretty sure we have an issue for that in the backlog somewhere

PR #424, opened 2024-03-01
fix: version if installed via 'go install'

avirtopeanu-ionos:
Moves versioning logic to a new `version` pkg.

Use `buildinfo` to get version, since this is empty if installed via `go install`.

## `ionosctl version` before
```
 % go install github.com/ionos-cloud/ionosctl/v6@(some-commit)
go: downloading github.com/ionos-cloud/ionosctl/v6 v...

 % ionosctl version
DEV-
```

## `ionosctl version` after
```
 % go install github.com/ionos-cloud/ionosctl/v6@67c7f98                                     
go: downloading github.com/ionos-cloud/ionosctl/v6 v6.7.5-0.20240301130018-67c7f986faf0

 % i version
DEV-67c7f98
```

## `ionosctl version` installed via build script (before & after - no change - ldflags are used)

```
 % i version
DEV-67c7f98
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -7,6 +7,14 @@ import (
 	"strings"
 )
 
+type Option func(Config) Config
+type Config struct {
+	ReadBuildInfo BuildInfoFetcher
+}
+
+// BuildInfoFetcher specifies a custom build info fetcher, only really exists for testing
+type BuildInfoFetcher func() (*debug.BuildInfo, bool)
+
 // Variables sent at link time via ldflags
 var (
 	Version   string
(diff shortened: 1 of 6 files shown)


internal/version/version.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Get returns the current CLI version, preferring the value set via ldflags
+func Get() string {
+	if Version != "" || GitCommit != "" {
+		return getVersionViaLdFlags()
+	}
+
+	info, ok := debug.ReadBuildInfo()
+	if ok {
+		versionParts := strings.Split(info.Main.Version, "-")
```
cbeti-ionos, line 25: I think that for something such as  
`go install github.com/ionos-cloud/ionosctl/v6@v6.7.2`
`go: downloading github.com/ionos-cloud/ionosctl/v6 v6.7.2`

the slice `versionParts` will get `["v6.7.2"]` value which will end up reporting an `unknown` version possibly.

avirtopeanu-ionos: Yeah, that might be the case, thanks! We can add an early return case to see if `info.Main.Version` is a semantic tag, in which case we could return it directly

cbeti-ionos (approved):
Looks good, version should be returned correctly now when using the typical build/installation methods.

avirtopeanu-ionos:
Force push because of rebase, added commits

[enhancement: add comment for version.Get](https://github.com/ionos-cloud/ionosctl/pull/424/commits/164a91fd13d6f97226747b6946a0aeab74814a69)
[doc: changelog](https://github.com/ionos-cloud/ionosctl/pull/424/commits/4b5534a2d6e68cf065eae0b54cf633614a35d1c9)
[fix: double v in interactive shell](https://github.com/ionos-cloud/ionosctl/pull/424/commits/1f69d079f02244bb430dc6b2f5d7dccb15e4e7e7)

PR #452, opened 2024-07-09
Update README.md

hkacmaz:
--token $IONOS_TOKEN dollar sign is needed for exposed variable


avirtopeanu-ionos (approved):
Hi, sure we can change this to have a dollar sign

I think in its original state it was meant just to be a place-holder for the flag's value, but it doesn't hurt to show that you could use the `IONOS_TOKEN` env var here.

PR #460, opened 2024-10-28
add cdn operations

digna-ionos:
## Checklist

- [ ] PR name added as appropriate (e.g. `feat:`/`fix:`/`doc:`/`test:`/`refactor:`)
- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Sonar Cloud Scan
- [ ] Jira task updated


commands/root.go
```diff
@@ -249,6 +250,9 @@ func addCommands() {
 			logging_service.LoggingServiceCmd(), constants.DefaultLoggingServiceApiURL,
 		),
 	)
+
+	// CDN
+	rootCmd.AddCommand(cdn.Command())
```
avirtopeanu-ionos, line 255: Default API Url is `https://cdn.de-fra.ionos.com/distributions` I think we need to add this with a changed default API URL:

 `	rootCmd.AddCommand(funcChangeDefaultApiUrl(cdn.Command(), constants.DefaultCDNApiURL))`

commands/cdn/distribution/list.go
```diff
@@ -0,0 +1,92 @@
...
+				req = req.FilterState(viper.GetString(fn))
+			}
+			if fn := core.GetFlagName(c.NS, constants.FlagCDNDistributionFilterDomain); viper.IsSet(fn) {
+				req = req.FilterDomain(viper.GetString(fn))
+			}
+			if fn := core.GetFlagName(c.NS, constants.FlagOffset); viper.IsSet(fn) {
+				req = req.Offset(viper.GetInt32(fn))
+			}
+			if fn := core.GetFlagName(c.NS, constants.FlagLimit); viper.IsSet(fn) {
+				req = req.Limit(viper.GetInt32(fn))
```
avirtopeanu-ionos, line 65: I know it would make sense to use `--limit` but sadly we already use `--max-results` for all other products, and changing the name for CDN will be confusing for the user

Perhaps in the future we could come up with a better pagination system

commands/cdn/distribution/routingrules/routing_rules.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/ionos-cloud/ionosctl/v6/internal/printer/json2table/jsonpaths"
+	"github.com/ionos-cloud/ionosctl/v6/internal/printer/jsontabwriter"
+	"github.com/ionos-cloud/ionosctl/v6/internal/printer/tabheaders"
+	cdn "github.com/ionos-cloud/sdk-go-cdn"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var (
+	allRoutingRulesColumns     = []string{"Scheme", "Prefix", "Host", "Caching", "Waf", "RateLimitClass", "SniMode", "GeoRestrictions"}
```
avirtopeanu-ionos, line 20: This might be just a nitpick and could be ignored...

But I would rename `GeoRestrictions` to reflect that it is a block list (whatever naming you want honestly, i.e. `GeoBlockList` or `BlacklistCountries` etc etc.)

Reasoning: I suspect the API might change in the future and they could add, for example, a whitelist

avirtopeanu-ionos:
Let's also add the changes to the changelog, as well as a bats test suite, this looks good!

commands/cdn/distribution/routingrules/routing_rules.go
digna-ionos: GeoRestrictions can contain allowList or blockList. The struct contains both

docs/subcommands/CLI Setup/logout.md
```diff
@@ -26,7 +26,7 @@ If any such fields are found, their values will be replaced with an empty string
 AUTHENTICATION ORDER
 ionosctl uses a layered approach for authentication, prioritizing sources in this order:
   1. Global flags
-  2. Environment variables
+  2. Environment variables credentials test failed
```
avirtopeanu-ionos, line 29: Did  this generate like this?

`credentials test failed`

digna-ionos: seems so

commands/cdn/distribution/create.go
```diff
@@ -0,0 +1,113 @@
...
+	cdn "github.com/ionos-cloud/sdk-go-cdn"
+
+	"github.com/ionos-cloud/ionosctl/v6/internal/client"
+	"github.com/spf13/viper"
+
+	"github.com/ionos-cloud/ionosctl/v6/internal/core"
+)
+
+func Create() *core.Command {
+	cmd := core.NewCommand(context.Background(), nil, core.CommandBuilder{
```
avirtopeanu-ionos, line 22: Is there a reason why we not to use `NewCommandWithJsonProperties` ?

The user might not know exactly what json we expect here for `--routing-rules`.

NewCommandWithJsonProperties adds two flags `--json-properties` and `--json-properties-example`. The latter if set, it prints an example json (which for example could be piped to a .json file), and exits, so the user can use that as a sort of guide

commands/cdn/distribution/dsitribution.go
avirtopeanu-ionos, line 1: misspelled file `dsitribution` xD

commands/cdn/distribution/create.go
glimberea, line 22: I believe it might actually be more appropriate to have the `create` command use a json properties file to create the object. From what I can tell, the routing rules have quite a complex structure and it might be easier for the user to just use a properties file.

commands/cdn/distribution/routingrules/routing_rules.go
glimberea: it might actually be more legible to split this column into 2, one for `allowList` and one for `blockList`, imho

commands/cdn/distribution/create.go
digna-ionos: JsonProperties is kind of vague. Having the custom name is more unambiguous. We can have a routing-rules example if we want.

avirtopeanu-ionos: sure

internal/constants/constants.go
```diff
@@ -78,6 +78,13 @@ Within each layer, a token takes precedence over a username and password combina
 	FlagLoggingPipelineLogType          = "log-type"
 	FlagLoggingPipelineLogRetentionTime = "log-retention-time"
 
+	FlagCDNDistributionFilterDomain  = "filter.domain"
+	FlagCDNDistributionFilterState   = "filter.state"
```
avirtopeanu-ionos: This might be a nitpick, but using dots on flags is not standard and might be unexpected by users

I think it does work with Cobra but why not keep them as `--domain` and `--state` like we do with other commands, for consistency sake?

digna-ionos: we can keep them like that. I've just kept them as they were in the api.

avirtopeanu-ionos (approved):

PR #489, opened 2025-01-27
fix: mongo return early if no snapshots

avirtopeanu-ionos:
After #472 , listing snapshots causes an error if there are no snapshots to list

Fixes this bug by returning early.

Did not add to changelog as this bug hadn't been released yet


avirtopeanu-ionos:
fixed in #485

Issue #505, opened 2025-03-11
`container-registry token scope delete` does not work


jplitza:
## Description

The subcommand `container-registry token scope delete` is not really usable. Its help says to specify the scope to delete with `--name`, but it actually only accepts a `--scope-id`, which then results in a stacktrace (see below).

## Expected behavior

The invocation `$ ionosctl container-registry token scope delete --registry-id ... --token-id ... --name 'path/*'` should remove the specified scope from the specified token in the specified registry.

## Environment

Go Version:
```
```
Ionosctl version:
```
6.8.1
```
OS:
```
Ubuntu 24.04
```
Shell:
```
bash
```

## How to Reproduce

```
ionosctl container-registry token scope delete --registry-id ... --token-id ... --scope-id 0
```

## Error and Debug Output

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xfee174]

goroutine 1 [running]:
github.com/ionos-cloud/ionosctl/v6/commands/container-registry/token/scopes.CmdGetTokenScopesDelete(0xc000380200)
        github.com/ionos-cloud/ionosctl/v6/commands/container-registry/token/scopes/delete.go:122 +0x7f4
github.com/ionos-cloud/ionosctl/v6/internal/core.NewCommand.func2(0xc0005cc608, {0x131e004?, 0x4?, 0x131e010?})
        github.com/ionos-cloud/ionosctl/v6/internal/core/command_runner.go:77 +0x102
github.com/spf13/cobra.(*Command).execute(0xc0005cc608, {0xc00015a050, 0x6, 0x6})
        github.com/spf13/cobra@v1.8.1/command.go:985 +0xaaa
github.com/spf13/cobra.(*Command).ExecuteC(0x1dd2660)
        github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.8.1/command.go:1041
github.com/ionos-cloud/ionosctl/v6/commands.Execute()
        github.com/ionos-cloud/ionosctl/v6/commands/root.go:54 +0x1e
main.main()
        ./main.go:8 +0xf
```

avirtopeanu-ionos:
Hi! Sorry for the late reply, we will prioritize this and fix this soon!

PR #511, opened 2025-04-04
fix: various container-registry fixes

avirtopeanu-ionos:
- Added missing container-registry columns
  - The garbage collection days and time are now randomly generated to be a random working day during hours 10:00 - 16:00
  - Fixed listing container-registry token scopes
  - Fixed various panics related to nil ExpiryDate on container-registry tokens including #505

### DIFF
--- a/internal/printer/json2table/resource2table/container-registry.go
+++ b/internal/printer/json2table/resource2table/container-registry.go
@@ -82,9 +82,9 @@ func ConvertContainerRegistryVulnerabilityToTable(vulnerability containerregistr
 	return convertedVulnerability, nil
 }
 
-func ConvertContainerRegistryTokenScopesToTable(scopes []containerregistry.Scope) ([]map[string]interface{}, error) {
-	if scopes == nil || len(scopes) == 0 {
-		return nil, nil // empty output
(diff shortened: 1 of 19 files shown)


internal/printer/json2table/resource2table/container-registry.go
```diff
@@ -80,3 +81,21 @@ func ConvertContainerRegistryVulnerabilityToTable(vulnerability containerregistr
 
 	return convertedVulnerability, nil
 }
+
+func ConvertContainerRegistryTokenScopesToTable(scopes []containerregistry.Scope) ([]map[string]interface{}, error) {
+	if scopes == nil {
+		return nil, fmt.Errorf("could not retrieve Container Registry Token Scopes")
+	}
```
cbeti-ionos, line 88: if scopes is nil it returns error, but if scopes is empty it returns 0 value map, 
is this expected? 
wouldn't early return for both work ?
error message says "could not retrieve" but the function naming just implies conversion

avirtopeanu-ionos: `wouldn't early return for both work ?` yes sure, I'll change this. But we explicitly want an empty output if there are no token scopes, so I'll make it return a 0 value map.

This will also get rid of the confusing error (which honestly I just copy pasted from another similar resource)

cbeti-ionos (approved):

PR #626, opened 2025-12-22
fix: Set cluster version as default for nodepool creation

cavramoniu-ionos:
Changed nodepool default version to be the same as the cluster version on nodepool creation.


glimberea (approved):
Looks great 👍🏻

PR #627, opened 2025-12-29
feature: Add GPUs column to template list/get commands

cavramoniu-ionos:
Added the GPUs column to the template list/get commands and the auto-completion for the --template-id flag.

### DIFF
--- a/internal/printer/json2table/resource2table/cloudapi.go
+++ b/internal/printer/json2table/resource2table/cloudapi.go
@@ -2,6 +2,8 @@ package resource2table
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/printer/json2table/resource2table/cloudapi.go
```diff
@@ -173,3 +173,71 @@ func ConvertRequestToTable(request ionoscloud.Request) ([]map[string]interface{}
...
+
+	result := ""
+	first := true
+	for model, count := range gpuMap {
+		if !first {
+			result += ", "
+		}
+		result += fmt.Sprintf("%dx %s", count, model)
+		first = false
+	}
```
glimberea: Assuming you employ the changes mentioned above, you can use the `strings.Join()` function to join the elements of the resulting slice into one string, with a separator between them.

PR #659, opened 2026-03-30
Feat: object storage bucket, policy, object, acl, version crud

cristiGuranIonos:
Add ionosctl object-storage command tree for IONOS S3-compatible Object Storage                                                             
Also adds S3 credential resolution with provenance display (env vars, config file)


commands/cfg/whoami.go
```diff
@@ -73,6 +73,29 @@ func handleProvenance(c *core.CommandConfig, cl *client.Client, authErr error) e
...
+
+	b.WriteString("  Secret Key:\n")
+	for i, src := range client.ObjectStorageSecretKeyOrder {
+		if skSrc == src {
+			b.WriteString(fmt.Sprintf("  * [%d] %s (USED)\n", i+1, src))
+		} else {
+			b.WriteString(fmt.Sprintf("    [%d] %s\n", i+1, src))
+		}
+	}
+
```
avirtopeanu-ionos, line 97: currently, this prints as:

```
Authentication layers, in order of priority:
* [1] environment variable: IONOS_TOKEN (USED)
  [2] environment variables: IONOS_USERNAME, IONOS_PASSWORD
  [3] credentials from config file: token
  [4] credentials from config file: username, password

Object Storage credential sources:
  Access Key:
    [1] environment variable: IONOS_S3_ACCESS_KEY
    [2] Object Storage access key from config file: s3AccessKey
  Secret Key:
    [1] environment variable: IONOS_S3_SECRET_KEY
    [2] Object Storage secret key from config file: s3SecretKey

```

To match the existing style, what do you say if we:
1. keep both ACCESS_KEY and SECRET_KEY as pulled from the same location (i.e. either both from env - or both from config)? 
2. keep all list items starting with non-capital letters ?
3. keep the header text similar (this is more of a nit pick though)

For example:
```
Authentication layers, in order of priority:
* [1] environment variable: IONOS_TOKEN (USED)
  [2] environment variables: IONOS_USERNAME, IONOS_PASSWORD
  [3] credentials from config file: token
  [4] credentials from config file: username, password

For Object Storage, in order of priority:
  [1] environment variables: IONOS_S3_ACCESS_KEY, IONOS_S3_SECRET_KEY
* [2] credentials from config file: s3AccessKey, s3SecretKey (USED)
```

commands/object-storage/bucket/head.go
```diff
@@ -0,0 +1,71 @@
...
+		PreCmdRun: func(c *core.PreCommandConfig) error {
+			return core.CheckRequiredFlags(c.Command, c.NS, constants.FlagName)
+		},
+		CmdRun: func(c *core.CommandConfig) error {
+			name := viper.GetString(core.GetFlagName(c.NS, constants.FlagName))
+
+			_, err := client.MustObjectStorage().ObjectStorageClient.BucketsApi.HeadBucket(c.Context, name).Execute()
+			if err != nil {
+				return err
+			}
```
avirtopeanu-ionos, line 46: [nitpick]

is it possible to wrap errors and describe what went wrong?

```
❯ i object-storage bucket head --name asdf
Error: 404 Not Found: 
```

commands/object-storage/bucket/list.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func ListBucketsCmd() *core.Command {
+	cmd := core.NewCommand(context.Background(), nil, core.CommandBuilder{
+		Namespace: "object-storage",
+		Resource:  "bucket",
+		Verb:      "list",
+		Aliases:   []string{"ls"},
+		ShortDesc: "List all contract-owned buckets",
+		Example:   "ionosctl object-storage bucket list\nionosctl object-storage bucket list --region eu-central-3",
```
avirtopeanu-ionos: here we use `--region` in the helptext but we actually expect `--location`

commands/object-storage/object/list.go
```diff
@@ -0,0 +1,144 @@
...
+	{Name: "Key", JSONPath: "Key", Default: true},
+	{Name: "Size", JSONPath: "Size", Default: true},
+	{Name: "LastModified", JSONPath: "LastModified", Default: true},
+	{Name: "StorageClass", JSONPath: "StorageClass", Default: true},
+	{Name: "ETag", JSONPath: "ETag"},
+}
+
+type listObjectInfo struct {
+	Key          string    `json:"Key"`
+	Size         int32     `json:"Size"`
```
avirtopeanu-ionos: is an int32 enough ? Is Size returned in Bytes or something else?

It would also be nice to keep it consistent with other CLI commands and add the unit after the returned number.

commands/object-storage/bucket/create.go
```diff
@@ -0,0 +1,53 @@
...
+			cfg.SetLocationConstraint(location)
+
+			req := client.MustObjectStorage().ObjectStorageClient.BucketsApi.CreateBucket(c.Context, name).
+				CreateBucketConfiguration(*cfg).XAmzBucketObjectLockEnabled(objectLock)
+			_, err := req.Execute()
+			if err != nil {
+				return err
+			}
+
+			fmt.Fprintf(c.Command.Command.OutOrStdout(), "Bucket %q created successfully in region %q\n", name, location)
```
avirtopeanu-ionos: most people who write ionosctl scripts may be used to `create` commands working with `-o json`.

However, this may fail depending on what query they may run

Do you think it is feasible to also do a `get` on the created bucket and return a table print which also supports -o json?

commands/object-storage/bucket/list.go
```diff
@@ -0,0 +1,64 @@
...
+
+				bi := bucketInfo{
+					Name:         b.GetName(),
+					CreationDate: b.GetCreationDate(),
+				}
+
+				loc, apiResp, locErr := client.MustObjectStorage().ObjectStorageClient.BucketsApi.GetBucketLocation(c.Context, b.GetName()).Execute()
+				if locErr != nil {
+					c.Verbose("failed to get location for bucket %q: %v", b.GetName(), locErr)
+					bi.Region = fmt.Sprintf("<%s>", apiResp.Status)
```
avirtopeanu-ionos: could apiResp be nil if we have an `locErr` ?

This may panic

internal/client/objectstorageclient.go
```diff
@@ -0,0 +1,201 @@
...
+
+// ResolveObjectStorageCredentials resolves Object Storage access and secret keys, tracking their source.
+// Credentials are resolved in priority order:
+//  1. Environment variables IONOS_S3_ACCESS_KEY / IONOS_S3_SECRET_KEY
+//  2. s3AccessKey / s3SecretKey in the current ionosctl config profile
+func ResolveObjectStorageCredentials() (accessKey, secretKey string, akSrc ObjectStorageAccessKeySource, skSrc ObjectStorageSecretKeySource, err error) {
+	src, cfgErr := retrieveConfigFile()
+	if cfgErr != nil {
+		return accessKey, secretKey, akSrc, skSrc, fmt.Errorf("failed to retrieve config file: %w", cfgErr)
+	}
```
avirtopeanu-ionos: ok, we may fail to retrieve the config file - but what if the user has valid env vars? shouldn't we continue gracefully with those?

commands/object-storage/bucket/create.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func CreateBucketCmd() *core.Command {
+	cmd := core.NewCommand(context.Background(), nil, core.CommandBuilder{
+		Namespace: "object-storage",
+		Resource:  "bucket",
+		Verb:      "create",
+		Aliases:   []string{"c"},
+		ShortDesc: "Create a contract-owned bucket",
+		Example:   "ionosctl object-storage bucket create --name my-bucket\nionosctl object-storage bucket create --name my-bucket --region eu-central-3\nionosctl object-storage bucket create --name my-bucket --object-lock",
```
avirtopeanu-ionos:
```suggestion
		Example:   "ionosctl object-storage bucket create --name my-bucket\nionosctl object-storage bucket create --name my-bucket --location eu-central-3\nionosctl object-storage bucket create --name my-bucket --object-lock",
```

avirtopeanu-ionos (approved):
nice work!

PR #669, opened 2026-05-05
feat: global wait

avirtopeanu-ionos:
Global `--wait` (`-w`) flag. Blocks until the API request completes and the resource (plus all parent resources) reaches a terminal ready state. Replaces per-command `--wait-for-request`/`--wait-for-state`/`--wait-for-deletion` which are now deprecated (still work, show deprecation warning).

### How it works

1. **URL capture** (during command execution): All SDK HTTP clients are wrapped with a transport that captures the HTTP method, request URL, and `Location` header from POST/PUT/PATCH/DELETE responses. When the command prints output, a `BeforeRender` hook extracts the resource URL - either from the response body's `href` field (CloudAPI), or by appending the response `id` to the transport-captured collection URL (MariaDB, Postgres, Mongo). The hook suppresses output so it can be re-rendered after waiting.

2. **Polling** (after command returns): First polls the request status URL (from the `Location` header, e.g. `/requests/{id}/status`) until DONE, this ensures the API request itself has completed. Then polls the resource URL, checking `metadata.state` or `metadata.status` for a terminal state (AVAILABLE, ACTIVE, READY, DONE, INACTIVE, SUSPENDED). Walks up the URL path hierarchy to also poll each parent resource (e.g. server -> datacenter). For action commands (start/stop/reboot/suspend/resume) that don't support GET on their endpoint, only the request status is polled. 404 during delete = success; 404 during create = transient retry. 429 respects `Retry-After`. Other 4xx/5xx fail immediately.

3. **Re-render** (after polling completes): Re-fetches the resource (respecting `--depth`) and re-renders the table output showing the final AVAILABLE state instead of the initial BUSY state. For JSON output, produces a clean GET response. Progress spinner shown on stderr for text output, suppressed for JSON.

After capture, polls the resource URL and walks up the path hierarchy to poll each parent resource too. Checks `metadata.state` or `metadata.status` for terminal states:
- **Success**: AVAILABLE, ACTIVE, READY, DONE, INACTIVE, SUSPENDED
- **Failure**: FAILED, ERROR, DESTROYING

404 during delete is treated as success (resource gone); 404 during create/update is treated as transient and retried. Other 4xx errors (400, 405, 409) fail immediately. 429 respects `Retry-After` (capped at 60s, context-aware).

### User impact

Old `-w` shorthand (was `--wait-for-request`) now means `--wait`. The `-W` shorthand (was `--wait-for-state`) has been removed due to conflict with `--weight` in NLB/targetgroup commands. Old long-form flags are hidden from help but still functional. Default timeout changed from 60s to 600s (10 minutes).

### Breaking changes

- `-t` shorthand removed from `--token-id` (container-registry) and `--type` (dns record create) to avoid conflict with global `--timeout` (`-t`). Use the long form instead.
- `-W` shorthand no longer available.

### Known limitations

- `--wait` combined with `--all` (bulk delete) only waits for the last deleted resource. For guaranteed completion, delete resources individually with `--wait`.


internal/waitfor/wait.go
```diff
@@ -28,13 +28,11 @@ type InterrogateRequestFunc func(c *core2.CommandConfig, requestId string) (stat
 
 // WaitForRequest waits for Request to be executed
 func WaitForRequest(c *core2.CommandConfig, interrogator InterrogateRequestFunc, requestId string) error {
```
glimberea: let's delete unused legacy functions, please

internal/constants/constants.go
```diff
@@ -358,7 +357,7 @@ const (
 	ArgServerUrlShort = "u"
 	ArgToken          = "token"
 	FlagTokenId       = "token-id"
-	ArgTokenShort     = "t"
+	ArgTokenShort     = ""
```
glimberea, line 360: this affects the `ionosctl token` and `ionosctl login` commands as well, it should be documented

internal/globalwait/globalwait.go
```diff
@@ -0,0 +1,864 @@
...
+func WaitForAvailable(w io.Writer, token, username, password string) error {
+	href := getHref()
+	if href == "" {
+		return nil
+	}
+
+	timeoutSec := viper.GetInt(constants.ArgTimeout)
+	if timeoutSec <= 0 {
+		if timeoutSec == 0 {
+			fmt.Fprintf(w, "Warning: --timeout 0 is not supported, using default %ds\n", constants.DefaultTimeoutSeconds)
```
glimberea: warning message should be for all values `<= 0`, not jut `== 0`

commands/root.go
```diff
@@ -149,10 +163,59 @@ func init() {
...
+		rootPFlagSet.MarkDeprecated(old, "use --wait instead")
+	}
+
+	// Deprecated -W shorthand removed: conflicts with --weight (-W) in NLB/targetgroup commands.
+	// The long form --wait-for-state (above) still works. -W was rarely used directly.
+	rootPFlagSet.Bool("wait-for-state-deprecated", false, "DEPRECATED: use --wait instead")
+	_ = viper.BindPFlag("wait-for-state-deprecated", rootPFlagSet.Lookup("wait-for-state-deprecated"))
+	rootPFlagSet.MarkHidden("wait-for-state-deprecated")
+	rootPFlagSet.MarkDeprecated("wait-for-state-deprecated", "use --wait (-w) instead")
+
```
glimberea: is this left in by mistake? I fail to see the need for this piece of code

internal/constants/constants.go
avirtopeanu-ionos: it is documented in the changelog - do you think we should add this somewhere else?

glimberea: No, changelog should be sufficient. I must've not seen it when I first looked over it. Thanks!

commands/root.go
avirtopeanu-ionos: removed, thanks!

internal/globalwait/globalwait.go
avirtopeanu-ionos: fixed, thanks!

internal/waitfor/wait.go
avirtopeanu-ionos: done, thanks! Also I have done a little refactor which made --promote-volume use the globalwait pkg - so now we can remove the whole old package.

glimberea (approved):

PR #680, opened 2026-07-01
feat: add volume performance classes to volume --type completion

mimihalescu:
Adds the new volume performance classes `ESSENTIAL`, `BALANCED` and `PERFORMANCE` to the `volume create --type` shell completion. Also fixes a pre-existing bug where this completion was registered on `--licence-type` instead of `--type`, so `--type` had no completion at all.

Draft until the classes are live on the API.

### DIFF
--- a/commands/compute/volume/create.go
+++ b/commands/compute/volume/create.go
@@ -51,8 +51,8 @@ ionosctl compute volume create --datacenter-id DATACENTER_ID --name NAME --image
 	})
 	cmd.AddSetFlag(cloudapiv6.ArgLicenceType, "", "LINUX", constants.EnumLicenceType, "Licence Type of the Volume")
 	cmd.AddStringFlag(constants.FlagType, "", "HDD", "Type of the Volume")
-	_ = cmd.Command.RegisterFlagCompletionFunc(cloudapiv6.ArgLicenceType, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"HDD", "SSD", "SSD Standard", "SSD Premium"}, cobra.ShellCompDirectiveNoFileComp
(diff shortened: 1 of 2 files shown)


commands/compute/volume/create.go
```diff
@@ -51,8 +51,8 @@ ionosctl compute volume create --datacenter-id DATACENTER_ID --name NAME --image
 	})
 	cmd.AddSetFlag(cloudapiv6.ArgLicenceType, "", "LINUX", constants.EnumLicenceType, "Licence Type of the Volume")
 	cmd.AddStringFlag(constants.FlagType, "", "HDD", "Type of the Volume")
-	_ = cmd.Command.RegisterFlagCompletionFunc(cloudapiv6.ArgLicenceType, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"HDD", "SSD", "SSD Standard", "SSD Premium"}, cobra.ShellCompDirectiveNoFileComp
+	_ = cmd.Command.RegisterFlagCompletionFunc(constants.FlagType, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"HDD", "SSD", "SSD Standard", "SSD Premium", "ESSENTIAL", "BALANCED", "PERFORMANCE"}, cobra.ShellCompDirectiveNoFileComp
```
avirtopeanu-ionos: If the API team means to have most of the users use the new types, I would move them in front of the HDD / SSD choices, else 90+% of the users will use HDD / SSD


```suggestion
		return []string{"ESSENTIAL", "BALANCED", "PERFORMANCE", "HDD", "SSD", "SSD Standard", "SSD Premium"}, cobra.ShellCompDirectiveNoFileComp
```

